PR #3, opened 2019-11-22
Adding file format as a CSS class to allow specific CSS styling

akosma:
This PR adds the "format" parameter as a CSS class, so that different styling can be used for different output styles.

This helps in the case of browsers not properly rendering SVG images coming from a third-party; for example, SVG images produced in the Ditaa format appear normally on Antora sites, while SVG images produced by Blockdiag do not (tested in Chrome and Firefox) as they appear with zero width and zero height.

Having a separate CSS class would allow to add some additional parameters to fix the display of the image.

### DIFF
--- a/src/asciidoctor-kroki.js
+++ b/src/asciidoctor-kroki.js
@@ -43,14 +43,23 @@ const processKroki = (processor, parent, attrs, diagramType, diagramText, contex
   if (subs) {
     diagramText = parent.applySubstitutions(diagramText, parent.$resolve_subs(subs))
   }
-  const role = attrs.role
   const blockId = attrs.id
   const title = attrs.title
   const target = attrs.target
   const format = attrs.format || 'svg'
+  let role = attrs.role
+  if (role) {
+    if (format) {
+      role = `${role} ${format} kroki`
+    } else {
+      role = `${role} kroki`
+    }
+  } else {
+    role = 'kroki'
+  }
   const imageUrl = createImageSrc(doc, diagramText, diagramType, target, format, context.vfs)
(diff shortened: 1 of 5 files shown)


src/asciidoctor-kroki.js
```diff
@@ -43,14 +43,23 @@ const processKroki = (processor, parent, attrs, diagramType, diagramText, contex
...
   }
-  const role = attrs.role
   const blockId = attrs.id
   const title = attrs.title
   const target = attrs.target
   const format = attrs.format || 'svg'
+  let role = attrs.role
+  if (role) {
+    if (format) {
+      role = `${role} ${format} kroki`
```
Mogztter: I feel like the class name "png" or "svg" is too broad, what about `kroki-format-${format}`?

Mogztter:
> This helps in the case of browsers not properly rendering SVG images coming from a third-party; for example, SVG images produced in the Ditaa format appear normally on Antora sites, while SVG images produced by Blockdiag do not (tested in Chrome and Firefox) as they appear with zero width and zero height.

Ideally, the Kroki server should normalized the SVG produced by these diagrams libraries.
The Asciidoctor Kroki extension could then set the height and width (if these attributes are defined).

In the meantime, I think it can be useful to add the format as a CSS class.

src/asciidoctor-kroki.js
akosma: Works for me! You are right

akosma:
I've pushed changes to this, let me kno!

akosma:
:+1:

PR #8, opened 2019-12-20
resolves #7 creates a literal block when format is txt, utxt or atxt

### LINKED ISSUE #7: Create a literal block if the output format is txt or utxt (unicode plain text)
Reference: https://github.com/yuzutech/kroki/issues/38

Mogztter:


Mogztter:
@akosma It's working fine when using `utxt`. 

kroki-manpage


With ASCII only text spaces are not properly escaped/preserved:

kroki-manpage-off


And here's the code to generate the manpage:

```js
const asciidoctor = require('@asciidoctor/core')()
require('@asciidoctor/manpage-converter')()
const kroki = require('asciidoctor-kroki')
kroki.register(asciidoctor.Extensions)

const options = {
  attributes: { backend: 'manpage', doctype: 'book' },
  standalone: true
}
const input = `= kroki(1)
:doctype: manpage

== Name

kroki - creates diagrams from textual descriptions

== Synopsis

[plantuml,format=utxt]
----
Text->Diagram:convert
----`

const manpage = asciidoctor.convert(input, options)
console.log(manpage)
```

akosma:
Whoa! Excellent thanks!

PR #55, opened 2020-04-28
Issue 53 inline referenced vegalite data file

stenzengel:
@Mogztter Could you please have a look at this proposition? Feel free to modify as you like. In addition to fixing the issue, I added a new dependency to "json5". If you do not want this dependency, revert commit "use json5". I also added minimal editor settings in folder ".vscode" which maintains your formatting and runs mocha tests in VS Code. This could simplify participation in the future. If you do not want this, it would help me if you replace them with an entry in ".gitignore". Furthermore, I added the comment "// @ts-check" at the beginning of "asciidoctor-kroki.js" which helps me to avoid silly errors.


src/asciidoctor-kroki.js
```diff
@@ -97,6 +101,46 @@ const processKroki = (processor, parent, attrs, diagramType, diagramText, contex
   }
 }
 
+/**
+ * @param {string} diagramText
+ * @param { {vfs: { read: function} } } context
+ * @returns {string}
+ */
+function preprocessVegaLite (diagramText, context) {
```
Mogztter, line 109: We should probably move this function into a dedicated file and export it. It would allow to write unit tests.

src/asciidoctor-kroki.js
```diff
@@ -1,3 +1,4 @@
+// @ts-check
```
Mogztter, line 1: What does this comment do exactly? Since we are not using TypeScript we should probably use `@ts-nocheck` no?

package.json
```diff
@@ -27,6 +27,7 @@
     "browserify": "^16.3.0",
     "chai": "^4.2.0",
     "dirty-chai": "^2.0.1",
+    "json5": "^2.1.3",
```
Mogztter, line 30: I believe that it's a runtime dependency since we are using it in the runtime code.
We need to check if it's working in a browser environment.

src/asciidoctor-kroki.js
stenzengel: This comment enables type checking in JS files by the TypeScript compiler. JSDoc types are used for (optional) type information (https://www.typescriptlang.org/docs/handbook/type-checking-javascript-files.html). With this comment, the TypeScript compiler is enabled in VS Code and an error would be shown for example, if I forgot an argument when calling `preprocessVegaLite` or if I forgot to return a result from this function (which actually happened to me yesterday). This comment should not disturb in other editors. One could also generally activate (and configure) the TypeScript check for all JS files with the help of a `tsconfig.json` file and explicitly deactivate it if necessary.

src/asciidoctor-kroki.js
stenzengel: Good idea, I will have a look at it tomorrow.

package.json
stenzengel: What's the easiest way to test, if it is working in a browser environment? 

By the way, I was not able to run the "test:browser", because of the following error:
```
Unable to run tests using Puppeteer Error: Failed to launch the browser process!
.../asciidoctor-kroki/node_modules/puppeteer/.local-chromium/linux-737027/chrome-linux/chrome: error while 
loading shared libraries: libXcursor.so.1: cannot open shared object file: No such file or directory
```

I'm working on Windows with WSL (Windows Subsystem for Linux), which I installed yesterday. I'll probably have to install something there, but I haven't had time to have a closer look at it. The other NPM scripts worked fine.

src/asciidoctor-kroki.js
Mogztter: Ok I was asking because you said "...which helps me to avoid silly errors.". So I thought that you  would have needed to disable TypeScript compiler.
But I'm fine with this comment.

package.json
Mogztter: > What's the easiest way to test, if it is working in a browser environment?

We have tests running with Puppeteer/Chrome to make sure that it's working in a browser environment.
You can also build the project and open the page https://github.com/Mogztter/asciidoctor-kroki/blob/master/test/browser/index.html using a local HTTP server (ie. it won't work if you open the page in your browser using the `file://` protocol because CORS)

> By the way, I was not able to run the "test:browser", because of the following error:

Puppeteer/Chrome requires a few system dependencies so you will probably need to install `libxcursor1`. Are you using Ubuntu on WSL? If so you can try: https://github.com/puppeteer/puppeteer/issues/807#issuecomment-436588786

stenzengel: I moved JSON5 from "devDependencies" to "dependencies". "test/browser/index.html" in browser and "npm run test:browser" work (Thanks for the hints!). Is there anything else to do regarding the new dependency?

src/asciidoctor-kroki.js
stenzengel: I moved the function to a new file and added some tests. There is a "TODO" in "test\preprocess.spec.js" concerning inlining referenced remote files. Could you please have a look at it?

stenzengel:
@Mogztter I'm not sure what is the best way to handle a parsing error of the referencing JSON or a read error of a non-existing referenced data file in the `catch` blocks in `src\preprocess.js`. Yesterday I thought that the pre-processing should not cause any additional problems and leave the error handling completely to Vega-Lite. I.e. the `preprocessVegaLite`  should only log problems and not throw any errors. This is the currently implemented code. Today I'm not sure anymore. We could follow a fail-fast strategy and simply rethrow the original error with additional info. Invalid JSON and a non-existing referenced file cannot be processed meaningfully by the server anyway and we could perhaps give a better error report to the user. What do you prefer?

Mogztter:
> I'm not sure what is the best way to handle a parsing error of the referencing JSON or a read error of a non-existing referenced data file in the catch blocks in src\preprocess.js. Yesterday I thought that the pre-processing should not cause any additional problems and leave the error handling completely to Vega-Lite. I.e. the preprocessVegaLite should only log problems and not throw any errors. This is the currently implemented code. Today I'm not sure anymore. We could follow a fail-fast strategy and simply rethrow the original error with additional info. Invalid JSON and a non-existing referenced file cannot be processed meaningfully by the server anyway and we could perhaps give a better error report to the user. What do you prefer?

I think there are different cases:

- handle a parsing error of the referencing JSON: we should probably abort since the file is invalid there's no point to send it over the wire
- a read error of a non-existing **local** data file: we should probably also abort right away if the local file cannot be resolved: `file://../path/to/file`, `file:///path/to/file`, `../path/to/file`, `/path/to/file`, `c:\\path\to\file`...
- a read error of a non-existing **remote** data file: this one is tricky because at some point we might add a whitelist and/or allow loading files from remote URI in Kroki (see: https://github.com/yuzutech/kroki/issues/60). The reason is that Kroki can be installed on your infrastructure so while it makes sense to disallow loading file from remote URI on a public instance, it can also makes sense to enable it on private instances. It could also makes sense to add trusted domains on the public instance (not sure about it yet). In other words, the remote file can be unavailable on your local machine (network, security...) but available on your private instance of Kroki. For now, we should probably just warn the user that a file cannot be resolved but send it to Kroki anyway. We could also add an option to configure when remote files should be resolved (but it seems too early).

package.json
Mogztter: No I think it's good.
I will try to checkout your branch in the next days to make sure that everything is working in the browser extension: https://github.com/asciidoctor/asciidoctor-browser-extension

test/preprocess.spec.js
```diff
@@ -0,0 +1,66 @@
...
+    expect(preprocessVegaLite(unexistingReferencedFile, {})).to.be.equal(unexistingReferencedFile)
+  })
+
+  it('should return diagramText with inlined local file referenced with "data.url"', () => {
+    const referencedLocalCsvFile = `{
+  "data": {
+    "url": "test/fixtures/vegalite-data.csv"
+  }
+}`
+    const inlinedLocalCsvFile = String.raw`{
```
Mogztter, line 37: Any reason to add `String.raw` here?

test/preprocess.spec.js
```diff
@@ -0,0 +1,66 @@
...
+}`
+    expect(preprocessVegaLite(referencedLocalCsvFile, {})).to.be.equal(inlinedLocalCsvFile)
+  })
+
+  // TODO "url" should be https://raw.githubusercontent.com/Mogztter/asciidoctor-kroki/master/test/fixtures/vegalite-data.csv
+  // or could be deleted
+  it('should return diagramText with inlined remote file referenced with "data.url"', () => {
+    const referencedRemoteCsvFile = `{
+  "data": {
+    "url": "https://raw.githubusercontent.com/stenzengel/asciidoctor-kroki/issue-53-inline-referenced-vegalite-data-file/test/fixtures/vegalite-data.csv"
```
Mogztter, line 53: I think we might need to use a mock to test all the cases, file not found 404, timeout, 200 with empty body, etc... 
In theory, it's straightforward (see: https://www.npmjs.com/package/mock-http-server) but in practice it won't work because we are using `unxhr` to do synchronous requests using `child_process`. 
In Asciidoctor.js, I'm running `mock-http-server` in a new Node.js process: https://github.com/asciidoctor/asciidoctor.js/pull/886/files#diff-5250906497228a192b5b8d58cfb7ed6bR9

But leave it for now as it would introduce too much changes at once. I will update the URL to `https://raw.githubusercontent.com/Mogztter/asciidoctor-kroki/master/test/fixtures/vegalite-data.csv` once it's merged.

src/preprocess.js
```diff
@@ -0,0 +1,46 @@
...
+    // Same code as in Vega-Lite:
+    // https://github.com/vega/vega-lite/blob/master/src/compile/data/source.ts
+    let type = /(?:\.([^.]+))?$/.exec(data.url)[1]
+    if (['json', 'csv', 'tsv', 'dsv', 'topojson'].indexOf(type) < 0) {
+      type = 'json'
+    }
+    data.format = { type: type }
+  }
+  data.url = undefined
+  return JSON.stringify(diagramObject, undefined, 2)
```
Mogztter, line 45: we should probably trim all the whitespaces as it will produce a shorter URL:

```suggestion
  return JSON.stringify(diagramObject)
```

test/preprocess.spec.js
stenzengel: It lets me write `\n` instead of `\\n` for the `values` property, because escapes are not processed in `String.raw`. This is easier for me to grasp.

Mogztter: Oh I see that's neat :ok_hand:

stenzengel:
I've just pushed my changes containing the requested changes concerning whitespace and error handling. Thanks for the useful feedback!

Mogztter:
> I've just pushed my changes containing the requested changes concerning whitespace and error handling. Thanks for the useful feedback!

Awesome, I will try to take some time in the next few days to checkout your branch and give it a try :smiley: 
If it's working fine, it will provide a good foundation to implement the same mechanism for PlantUML and Vega.

src/preprocess.js
```diff
@@ -0,0 +1,82 @@
...
+  return error
+}
+
+/**
+ * @param {string} string
+ * @returns {boolean}
+ */
+function isRemoteUrl (string) {
+  try {
+    const url = new URL(string)
```
Mogztter, line 77: It requires Node.js >= 10 because the class is was available on the global object before.

> v10.0.0 | The class is now available on the global object.

https://nodejs.org/api/url.html#url_class_url

I don't think it's a big deal since we don't want to support Node < 10, so we should just set the engine in the `package.json`:

```json
{ "engines" : { "node" : ">=10" } }
```

Mogztter:
Could you please set the `engines` in the `package.json` and rebase on master?
The code looks good, once it's done I will merge it!

Thanks for your contribution Winfried :+1:

stenzengel:
I've added `engines` and rebased. It was a pleasure working with you, Guillaume!

Mogztter:
And merged! Thank you 😄

PR #64, opened 2020-05-10
Use getBaseDir not base_dir document attribute.

### LINKED ISSUE #62: Path resolution incorrect if no output directory specified
I'm looking at updating the vscode extension to the latest version of `asciidoctor-kroki` as part of https://github.com/asciidoctor/asciidoctor-vscode/issues/301

My working document is (say) here:

`/media/mulhollandd/KINGSTON/operating_information/Operating_Information/architecture/Operating_Information_Architecture.adoc`

and the document sets:

```asciidoc
:imagesdir: media
:kroki-server-url: https://kroki.io
:kroki-fetch-diagram:
```

The extension does set the `base_dir` **API option** but not `imagesoutdir`, `outdir`, `to_dir` or `base_dir` document attributes as appear to be expected in `fetch.js`:

https://github.com/Mogztter/asciidoctor-kroki/blob/f9b9abfb64be2339bde78bb77c41339e88dfcbfd/src/fetch.js#L5-L9

Under these circumstances, my images turn up in `/home/mulhollandd/media/` which surprised me no end and it was only looking through the code which made me look :wink:

The cause is `fs.writeFileSync` in:

https://github.com/Mogztter/asciidoctor-kroki/blob/f9b9abfb64be2339bde78bb77c41339e88dfcbfd/src/node-fs.js#L7-L12

where `filePath` might be something like `media/ba38556d595f9b57735747d9de9ee77e7335cdb2.svg`.

The node `fs` module which uses the `fs.openSync` on this relative path and somewhat deep in node internals it appears that if there is only a relative path then perhaps it uses the user's home directory (or perhaps this is what vs code is using as the cwd, although I doubt it). The [node filesystem docs](https://nodejs.org/api/fs.html#fs_file_paths) don't help me very much. 

I think the problem is that the asciidoctor-kroki extension is incorrectly handling `base_dir` which is an API option, not a document attribute. Perhaps we should be using instead the document attribute `docdir`?

https://asciidoctor.org/docs/user-manual/#piping-content-through-the-cli

Looking at the (Ruby) code:

https://github.com/asciidoctor/asciidoctor/blob/8f838709af08cdd5c3f1c91e4a6bba54e87cca54/lib/asciidoctor/convert.rb#L12-L20

It would seem that if `docdir` is used then it is set base on  `base_dir` if it is specified.

https://github.com/asciidoctor/asciidoctor/blob/88bc891d6a0143f3a2142f41d29cabe3cb21fe3f/lib/asciidoctor/document.rb#L380-L390

So I think the correct fix would be to replace `baseDir` with `docDir` along the lines of:

```js
  const imagesOutputDir = doc.getAttribute('imagesoutdir')
  const outDir = doc.getAttribute('outdir')
  const toDir = doc.getAttribute('to_dir')
  const docDir = doc.getAttribute('docdir') || ''
  const imagesDir = doc.getAttribute('imagesdir') || ''
  let dirPath
  if (imagesOutputDir) {
    dirPath = imagesOutputDir
  } else if (outDir) {
    dirPath = path.join(outDir, imagesDir)
  } else if (toDir) {
    dirPath = path.join(toDir, imagesDir)
  } else {
    dirPath = path.join(docDir, imagesDir)
  }
```

but I haven't entirely convinced myself so I haven't submitted a PR.

Thoughts?

danyill:
Closes #62.

### DIFF
--- a/test/delDir.js
+++ b/test/delDir.js
@@ -1,0 +1,15 @@
+// Until recursive: true is a stable part of Node
+// See: https://stackoverflow.com/questions/18052762/remove-directory-which-is-not-empty
+
+const fs = require('fs')
+const Path = require('path')
+
+module.exports.deleteDirWithFiles = function (path) {
+  if (fs.existsSync(path)) {
+    fs.readdirSync(path).forEach((file) => {
+      const curPath = Path.join(path, file)
+      fs.unlinkSync(curPath)
+    })
+    fs.rmdirSync(path)
+  }
+}
--- a/test/test.js
+++ b/test/test.js
@@ -93,10 +93,12 @@ plantuml::test/fixtures/alice.puml[svg,role=sequence]
       const doc = asciidoctor.convertFile(file, { extension_registry: registry, safe: 'unsafe' })
       fs.unlinkSync(doc.getAttributes().outfile)
       const imageLocation = path.join(doc.base_dir, doc.getAttributes().imagesdir)
-      await fsPromises.readdir(imageLocation).then((files) => {
+      try {
+        const files = await fsPromises.readdir(imageLocation)
         expect(files).to.have.lengthOf(1)
+      } finally {
         delDir.deleteDirWithFiles(imageLocation)
-      })
+      }
     })
     it('should download and save an image to a local folder', () => {
       const input = `
--- a/test/test.js
+++ b/test/test.js
@@ -2,6 +2,8 @@
 const fs = require('fs')
 const rusha = require('rusha')
 const pako = require('pako')
+const delDir = require('./delDir')
+const path = require('path')
 const chai = require('chai')
 const sinon = require('sinon')
 const rimraf = require('rimraf')
@@ -83,7 +85,25 @@ plantuml::test/fixtures/alice.puml[svg,role=sequence]
       const hash = rusha.createHash().update(`https://kroki.io/plantuml/svg/${encode(file)}`).digest('hex')
       expect(html).to.contain(`<img src=".asciidoctor/kroki/diag-${hash}.svg" alt="diagram">`)
     })
-    it('should download and save an image to a local folder and target name', () => {
+    it('should create diagrams in imagesdir if kroki-fetch-diagram is set', function (done) {
+      const registry = asciidoctor.Extensions.create()
+      asciidoctorKroki.register(registry)
+      const file = `${__dirname}/fixtures/fetch/doc.adoc`
+      const doc = asciidoctor.convertFile(file, { extension_registry: registry, safe: 'unsafe' })
+      fs.unlinkSync(doc.getAttributes().outfile)
+      const imageLocation = path.join(doc.base_dir, doc.getAttributes().imagesdir)
+      fs.readdir(imageLocation, (err, files) => {
+        if (err) {
+          done(`Image location: ${imageLocation} does not exist`)
+        }
+        if (files.length === 1) {
+          // tidy up
+          delDir.deleteDirWithFiles(imageLocation)
+          done()
+        }
+      })
+    })
+    it('should download and save an image to a local folder', () => {
       const input = `
 :imagesdir: .asciidoctor/kroki
 
--- a/test/test.js
+++ b/test/test.js
@@ -1,5 +1,6 @@
 /* global describe it before */
 const fs = require('fs')
+const fsPromises = require('fs').promises
 const rusha = require('rusha')
 const pako = require('pako')
 const delDir = require('./delDir')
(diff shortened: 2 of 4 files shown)


Mogztter:
@danyill So `getBaseDir` actually fixes this issue?

test/delDir.js
```diff
@@ -0,0 +1,15 @@
+// Until recursive: true is a stable part of Node
```
Mogztter, line 1: It hurts every time you need to remove recursively in Node :disappointed:

test/test.js
```diff
@@ -83,6 +85,24 @@ plantuml::test/fixtures/alice.puml[svg,role=sequence]
...
     })
+    it('should create diagrams in imagesdir if kroki-fetch-diagram is set', function (done) {
+      const registry = asciidoctor.Extensions.create()
+      asciidoctorKroki.register(registry)
+      const file = `${__dirname}/fixtures/fetch/doc.adoc`
+      const doc = asciidoctor.convertFile(file, { extension_registry: registry, safe: 'unsafe' })
+      fs.unlinkSync(doc.getAttributes().outfile)
+      const imageLocation = path.join(doc.base_dir, doc.getAttributes().imagesdir)
+      fs.readdir(imageLocation, (err, files) => {
+        if (err) {
```
Mogztter, line 96: Are you sure that an error is thrown if the directory does not exist? I'm not familiar with this API maybe it returns an empty list?

test/test.js
```diff
@@ -83,6 +85,24 @@ plantuml::test/fixtures/alice.puml[svg,role=sequence]
...
       expect(html).to.contain(`<img src=".asciidoctor/kroki/diag-${hash}.svg" alt="diagram">`)
     })
+    it('should create diagrams in imagesdir if kroki-fetch-diagram is set', function (done) {
+      const registry = asciidoctor.Extensions.create()
+      asciidoctorKroki.register(registry)
+      const file = `${__dirname}/fixtures/fetch/doc.adoc`
+      const doc = asciidoctor.convertFile(file, { extension_registry: registry, safe: 'unsafe' })
+      fs.unlinkSync(doc.getAttributes().outfile)
+      const imageLocation = path.join(doc.base_dir, doc.getAttributes().imagesdir)
+      fs.readdir(imageLocation, (err, files) => {
```
Mogztter, line 95: I think we should use `fsPromises.readdir` https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fspromises_readdir_path_options

```suggestion
      const files = await fsPromises.readdir(imageLocation)
```

test/test.js
```diff
@@ -83,6 +85,24 @@ plantuml::test/fixtures/alice.puml[svg,role=sequence]
       const hash = rusha.createHash().update(`https://kroki.io/plantuml/svg/${encode(file)}`).digest('hex')
       expect(html).to.contain(`<img src=".asciidoctor/kroki/diag-${hash}.svg" alt="diagram">`)
     })
+    it('should create diagrams in imagesdir if kroki-fetch-diagram is set', function (done) {
```
Mogztter, line 88:
```suggestion
    it('should create diagrams in imagesdir if kroki-fetch-diagram is set', async () => {
```

test/test.js
```diff
@@ -83,6 +85,24 @@ plantuml::test/fixtures/alice.puml[svg,role=sequence]
...
+      asciidoctorKroki.register(registry)
+      const file = `${__dirname}/fixtures/fetch/doc.adoc`
+      const doc = asciidoctor.convertFile(file, { extension_registry: registry, safe: 'unsafe' })
+      fs.unlinkSync(doc.getAttributes().outfile)
+      const imageLocation = path.join(doc.base_dir, doc.getAttributes().imagesdir)
+      fs.readdir(imageLocation, (err, files) => {
+        if (err) {
+          done(`Image location: ${imageLocation} does not exist`)
+        }
+        if (files.length === 1) {
```
Mogztter, line 99: This should be an assertion:

```suggestion
        expect(files.length).to.equal(1)
```

Mogztter: We can also use:

```suggestion
        expect(files).to.have.lengthOf(1)
```

test/test.js
```diff
@@ -83,6 +85,24 @@ plantuml::test/fixtures/alice.puml[svg,role=sequence]
...
+          // tidy up
+          delDir.deleteDirWithFiles(imageLocation)
```
Mogztter, line 101: I think we should do it in a `finally` block

test/test.js
danyill: It did, but I have changed as you recommended below.

test/test.js
danyill: I couldn't quite figure out how to do this.

danyill:
@Mogztter I've tried to follow your suggestions, just the last one wasn't apparent to me. I've rebased to master.
Yes, `getBaseDir()` does fix this issue.

test/test.js
```diff
@@ -83,7 +86,19 @@ plantuml::test/fixtures/alice.puml[svg,role=sequence]
...
+      const registry = asciidoctor.Extensions.create()
+      asciidoctorKroki.register(registry)
+      const file = `${__dirname}/fixtures/fetch/doc.adoc`
+      const doc = asciidoctor.convertFile(file, { extension_registry: registry, safe: 'unsafe' })
+      fs.unlinkSync(doc.getAttributes().outfile)
+      const imageLocation = path.join(doc.base_dir, doc.getAttributes().imagesdir)
+      await fsPromises.readdir(imageLocation).then((files) => {
+        expect(files).to.have.lengthOf(1)
+        delDir.deleteDirWithFiles(imageLocation)
+      })
```
Mogztter, line 99:
```suggestion
      try {
        const files = await fsPromises.readdir(imageLocation)
        expect(files).to.have.lengthOf(1)
      } finally {
        delDir.deleteDirWithFiles(imageLocation)
      }
```

test/test.js
Mogztter: That's OK, I've added a commit if you want to have a look: https://github.com/Mogztter/asciidoctor-kroki/pull/64/commits/9325af6c6190bc3511c9c97a10cee4fbbbf3fb2f

Mogztter:
>  I've tried to follow your suggestions, just the last one wasn't apparent to me. I've rebased to master.

Thanks, the code is looking good :+1: 
I've added the try/finally block so we are good to go!

> Yes, getBaseDir() does fix this issue.

Perfect :ok_hand:

danyill:
Thanks for the support and assistance :+1:

Issue #42, opened 2020-04-17
Switch to POST if the URI is too long


Mogztter:
When using very large Vega specification file (when data is embed in the file), the public instance of Kroki can return the following error "414 Request-URI Too Large nginx/1.14.0 (Ubuntu)".

I think we should use a POST request to avoid this error if the URI is longer than 4096.

See: https://github.com/yuzutech/kroki-go/issues/11

eshepelyuk:
@Mogztter why not use POST by default if kroki server supports this ?

Mogztter:
My main concern is corporate proxy/firewall. I believe that it's more likely that a POST request will be blocked than a GET request but maybe that's not a good enough reason. 

In one of my previous job the corporate proxy was a nightmare so I might be biased 😂

eshepelyuk:
Hey, lets introduce an attribute for this ? 
Called `kroki-http-method` - with following values
* get - mean always use GET
* post - means always use POST
* adaptive - try to use GET, unless query string length is less than 4096 bytes, then fall back to POST

Mogztter:
That's a good idea 👍

Mogztter:
There are several challenges we need to address.

When using the option "inline" we delegate all the work to Asciidoctor.js (ie. read the image data from the specified URI and generate a data URI). In this case, Asciidoctor.js will use a GET request to read the image data from the specified URI... :disappointed: 

Another issue is that we are currently using the Virtual File System to read the image data from the specified URI but if we want to use a POST request then we need to explicitly use an HTTP client.

Mogztter:
It's actually possible to set the target as a data-uri: https://github.com/asciidoctor/asciidoctor/issues/3657#issuecomment-629700269 :tada:

eshepelyuk:
@Mogztter although issue is closed, the README is not updated :((

Mogztter:
Indeed it was automatically closed when I merged the pull request, reopening :wink:

Mogztter:
README updated!
I think we can close now? @eshepelyuk any objection?

eshepelyuk:
@Mogztter plz close

PR #75, opened 2020-05-16
add preprocess support for PlantUML "!include"

### LINKED ISSUE #49: PlantUML !include preprocessor?
I'm using lot of **!include** files in PlantUML diagrams to have clear sturcture and DRY.

```
== UML class diagram test with kroki
[plantuml, "diagram-classes", svg]
....
!include <src/style-presets.iuml>

class BlockProcessor
class DiagramBlock
class DitaaBlock
class PlantUmlBlock

BlockProcessor <|-- DiagramBlock
DiagramBlock <|-- DitaaBlock
DiagramBlock <|-- PlantUmlBlock
....
```
For local builds with plantuml.jar it works, but with remote server like PlantUML server or kroki it does not work. I understand that it is not wanted, because it will do preprocessing with knowlegde of PlantUML syntax and has to be updated etc.

Is there other solution to solve similar structural problems at client side?

anb0s:
Fixes #49 

see:
http://plantuml.com/en/preprocessing
http://plantuml.com/en/commons
    
implement and test:

- [x] `!include local-file-path` (throw an error if file not found)
- [x] `!include local-file-path-with-spaces`
- [x] `!include ... # comment`
- [x] `!include remote-file-url` (skip and warn if url not accessible)
- [x] `!include <std/lib/path>` (skip and warn)
- [x] `!include file!index`
- [x] `!include file!id`
- [x] `!includeurl url` (compatibility : same as `!include remote-file-url`)
- [x] `!includesub id` (compatibility : includes `!startsub id ... !endsub`)
- [x] recursive include
- [x] cyclic include loop protection (throw an error if detected)
- [x] remove block `/' ... '/` and line `' ... '` comments
- [x] `!include_many local-file-or-url` (compatibility : same as normal `!include ...`)
- [x] `!include_once local-file-or-url` (throw an error if file included multiple times)
- [x] `!include local-file-path-or-url` if included file has `@startuml ... @enduml` block(s) it should include only first block (index 0)


anb0s:
@Mogztter please review the first naive implementation ;)

src/preprocess.js
```diff
@@ -56,6 +56,45 @@ ${diagramText}
...
 
+// @ts-check
+/**
+ * @param {string} diagramText
+ * @param {any} context
+ * @returns {string}
+ */
+module.exports.preprocessPlantUML = function (diagramText, context) {
+  // see: http://plantuml.com/en/preprocessing
+  const RegExInclude = /^(\s*)!(include(?:sub)?)\s+(.+?)(?:!(\w+))?$/i
```
Mogztter, line 67: You can take the following (more complete) Java regular expression from Kroki: https://github.com/yuzutech/kroki/blob/3309c057f1dfbfcb6ea58c1a717119f93db4a4ad/server/src/main/java/io/kroki/server/service/Plantuml.java#L72

A few notes:
- We should ignore includes from the stdlib (since the stdlib is included in the PlantUML jar and available on the Kroki server) or at least we should not thrown an error: `!include <some/stdlib>`
- You can have a trailing comment `# this is a comment`.
- If we want to resolve include properly we will need to implement the same logic regarding the "subid" feature (ie. `!include file.puml!2` will include the 3rd document in a file `file.puml`. Please note that it's also possible to use a custom id:

> For example, if you use !include foo.txt!1, the second @startuml/@enduml block within foo.txt will be included.
You can also put an id to some @startuml/@enduml text block in an included file using @startuml(id=MY_OWN_ID) syntax and then include the block adding !MY_OWN_ID when including the file, so using something like !include foo.txt!MY_OWN_ID.

Basically we will need an ultra-simplified PlantUML parser to read the file and extract the `@startuml/@enduml` blocks. I think we can leave this feature for now and issue a warning.

src/preprocess.js
```diff
@@ -56,6 +56,45 @@ ${diagramText}
...
+    RegExInclude,
+    (match, ...args) => {
+      const prefix = args[0]
+      // const include = args[1].toLowerCase()
+      const url = args[2].trim()
+      // const sub = args[3]
+      let vfs = context.vfs
+      if (typeof vfs === 'undefined' || typeof vfs.read !== 'function') {
+        vfs = require('./node-fs')
+      }
```
Mogztter, line 79: I will try to come up with a solution to hide this idiom

src/preprocess.js
```diff
@@ -56,6 +56,45 @@ ${diagramText}
...
+      let data = line
+      try {
+        data = prefix + vfs.read(url)
```
Mogztter, line 82: `prefix` is the whitespaces before the `include` directive? I'm not sure we need to preserve them but we will need to read the PlantUML code base to know:

1. if whitespaces are significant (I believe the answer is no)
2. if whitespaces are preserved or not when PlantUML resolves `include` directives.

I also assume that the `include` directive cannot be used inline.

Mogztter:
> reinclude protection (recursive include loop)

That's a good point but with your current implementation we do not resolve include recursively right? I mean we only resolve the first level of `include` no?
We should add a test on this case:

```
main.puml -> !include style.puml -> !include color.puml
```

And we should read PlantUML code to see how they prevent recursive include loop.

src/preprocess.js
anb0s: Thanks!

I've changed this PR to WIP/draft and will add all the needed include features, because i've  faced problems with our PlantUML files in some projects (std-lib, recursive includes etc.) and cannot use this basic not complete version ;)

- use the more complete regex
- skip lines with `!include <std/lib/path>`
- add `!include file!index` and `!include file!id`
- add recursive include support and prevent infinite include loop
- i'm not sure about add `!includesub id` and `!includeurl url`, because we are not using it and it's deprecated, but may be for compatibility we should add this...

Mogztter: > i'm not sure about add !includesub id and !includeurl url, because we are not using it and it's deprecated, but may be for compatibility we should add this...

yes until PlantUML removes them completely we will need to support `includesub` and `includeurl`.

src/preprocess.js
anb0s: Yes it's before the `include` directive. You are right `include` directive cannot be used inline. I think whitespaces are not significant, i preserve them because my tests fail as the compared strings are different - so it's just easier to preserve them :)

anb0s:
> > reinclude protection (recursive include loop)
> 
> That's a good point but with your current implementation we do not resolve include recursively right? I mean we only resolve the first level of `include` no?
> We should add a test on this case:
> 
> ```
> main.puml -> !include style.puml -> !include color.puml
> ```
> 
> And we should read PlantUML code to see how they prevent recursive include loop.

Yes it's not implemented and is a hint for the future. I will add this feature and check how to prevent include loop.

src/preprocess.js
anb0s: removed the whitespaces (prefix)

Mogztter:
@anb0s I see that you are making progress! Let me know when I can review it again. I've merged #42 in anticipation because once every includes will be inlined we will probably exceed the URI limit more often.
Anyway, keep up the good work 👍

anb0s:
@Mogztter please review the changes :)

anb0s:
Some tests related to "HTTP client / Adaptive mode" and "Conversion" are now failing and it looks not related to my last commit - i'm sure it worked some hours before ;)

Mogztter:
> Some tests related to "HTTP client / Adaptive mode" and "Conversion" are now failing and it looks not related to my last commit - i'm sure it worked some hours before ;)

I've just released and published a new version of Kroki and some images are not exactly the same with the latest version. We need to update the expected output accordingly.

Mogztter:
@danyill just fixed the tests, please rebase on master.

src/preprocess.js
anb0s: I've adressed the issues, please review

src/preprocess.js
anb0s: Still use this additionally with exists function

anb0s:
I've I've thought about some issues not yet adressed:
- @Mogztter : you have metioned trailing comments: `# this is a comment` but i didn't found any documentation about it
- `!include` in [comments](https://plantuml.com/de/commons): e.g. single line `'!include my-file.iuml'` and multiline `/' ... !include my-file.iuml ... '/` are processed, but must be skipped
- including whole file that has `@startuml ... @enduml` block(s) --> check how original PlantUML handles them: remove them from file or just include the whole file and PlantUML will complane at server? _We are using always `*.iuml` files without any `@startuml..@enduml` blocks, so it works for us ;)_
- we just include one file multiple times except if's not included cyclic (prevent infinite include loop), how is about [new/old preprocessor](https://plantuml.com/de/preprocessing):
> !include now allows multiple inclusions : you don't have to use !include_many anymore
> ...
> Note that there is also a !include_once directive that raises an error if a file is included several times.

--> should we support `!include_many` and `!include_once` ?

Mogztter:
> should we support !include_many and !include_once ?

Good catch! I wasn't even aware of these two directives 🤯
I've checked and they are still supported in the latest version of PlantUML so, unfortunately, we need to support them as well 😬 

> you have mentioned trailing comments: # this is a comment but i didn't found any documentation about it

Yes, it's not documented but you can give it a try: http://www.plantuml.com/plantuml/uml/ZOz1ImCn48Nl-HLnlQnWDmezUQginwgbwbdIxD0Dp4x2p0JflpSkw2AYU7elWVUzDnjHprecD6UH0fO12gCTLRFSECV-hJj67SkX2718f43QXJIvVGoUX_GG8GEwVhtQeITzkTyw1WGluAvNL_N_jQyX7mcuoxN5djj-RJ-SCJjWITJcMMunUhcqpTOV09jg7hF6H1xj7KZWEB-Q2_E5QyEd-n5zXZ1ImdGy_G5wEed2rCei_sqtUy3lWtzTiI2MEPMm0HhgzUy0


> !include in comments: e.g. single line '!include my-file.iuml' and multiline /' ... !include my-file.iuml ... '/ are processed, but must be skipped

it should be relatively easy, I think that an include directive is only valid if it starts with 0 or more spaces, otherwise it won't get processed. So, the following are valid

```
   !include /path/to/something.puml
!include /path/to/something.puml
 !include /path/to/something.puml
```
But the following are not:

```
a   !include /path/to/something.puml
\!include /path/to/something.puml
 '!include /path/to/something.puml
```

Multi-line comments are more complex to handle because we will need to ignore content when we find `/'` until we find `'/`.

> including whole file that has @startuml ... @enduml block(s) --> check how original PlantUML handles them: remove them from file or just include the whole file and PlantUML will complane at server? We are using always *.iuml files without any @startuml..@enduml blocks, so it works for us ;)

I think it should work, PlantUML will ignore/remove them.

I will try to take some time this week to review this great contribution 👍

anb0s:
> > should we support !include_many and !include_once ?
> 
> Good catch! I wasn't even aware of these two directives 🤯
> I've checked and they are still supported in the latest version of PlantUML so, unfortunately, we need to support them as well 😬

OK, i've added them to the list.

> > you have mentioned trailing comments: # this is a comment but i didn't found any documentation about it
> 
> Yes, it's not documented but you can give it a try: 

OK, also added to the list, should be easy.

> > !include in comments: e.g. single line '!include my-file.iuml' and multiline /' ... !include my-file.iuml ... '/ are processed, but must be skipped
> 
> it should be relatively easy, I think that an include directive is only valid if it starts with 0 or more spaces, otherwise it won't get processed.
> Multi-line comments are more complex to handle because we will need to ignore content when we find `/'` until we find `'/`.

OK, i've added to the list and added test and implemented, it was easy with regex ;)

> > including whole file that has @startuml ... @enduml block(s) --> check how original PlantUML handles them: remove them from file or just include the whole file and PlantUML will complane at server? We are using always *.iuml files without any @startuml..@enduml blocks, so it works for us ;)
> 
> I think it should work, PlantUML will ignore/remove them.

OK, we should first check how PlantUML handles it.

> I will try to take some time this week to review this great contribution 👍

Thank You!

anb0s:
I've added the last features discussed and have now only one to decide or verify:

To be specified: `!include local-file-path-or-url` should remove `@startuml` and `@enduml` lines from included file and include only first block (index 0) or include all blocks as one?

anb0s:
hm, i've tried all the files i've created for tests with original PlantUML and found that `!include_once` is not an include guard  in the included file itself like i've implemented, but it's just like `!include_many my-file.iuml `and `!include my-file.iuml` you can use `!include_once my-file` and than it will produce error if used multiple times:

produces error:
```
@startuml
  !include_once test\fixtures\plantuml\style-general.iuml
  alice -> bob
  !include_once test\fixtures\plantuml\style-general.iuml
@enduml
```

no error:
```
@startuml
  !include_once test\fixtures\plantuml\style-general.iuml
  alice -> bob
  !include test\fixtures\plantuml\style-general.iuml
@enduml
```

That's should be easy to change...

anb0s:
> hm, i've tried all the files i've created for tests with original PlantUML and found that `!include_once` is not an include guard in the included file itself like i've implemented, but it's just like `!include_many my-file.iuml `and `!include my-file.iuml` you can use `!include_once my-file` and than it will produce error if used multiple times:
> 
> produces error:
> 
> ```
> @startuml
>   !include_once test\fixtures\plantuml\style-general.iuml
>   alice -> bob
>   !include_once test\fixtures\plantuml\style-general.iuml
> @enduml
> ```
> 
> no error:
> 
> ```
> @startuml
>   !include_once test\fixtures\plantuml\style-general.iuml
>   alice -> bob
>   !include test\fixtures\plantuml\style-general.iuml
> @enduml
> ```
> 
> That's should be easy to change...

Done!

anb0s:
> To be specified: `!include local-file-path-or-url` should remove `@startuml` and `@enduml` lines from included file and include only first block (index 0) or include all blocks as one?

Tested with original PlantUML and adapted requirement:
`!include local-file-path-or-url` if included file has `@startuml ... @enduml` block(s) it should include only first block (index 0)

anb0s:
> > To be specified: `!include local-file-path-or-url` should remove `@startuml` and `@enduml` lines from included file and include only first block (index 0) or include all blocks as one?
> 
> Tested with original PlantUML and adapted requirement:
> `!include local-file-path-or-url` if included file has `@startuml ... @enduml` block(s) it should include only first block (index 0)

Done!

anb0s:
From my point of view the important `!include` features are implemented and tested.

Please review...

src/asciidoctor-kroki.js
```diff
@@ -44,6 +44,8 @@ const processKroki = (processor, parent, attrs, diagramType, diagramText, contex
   }
   if (diagramType === 'vegalite') {
     diagramText = require('./preprocess.js').preprocessVegaLite(diagramText, context)
+  } else if (diagramType === 'plantuml') {
+    diagramText = require('./preprocess').preprocessPlantUML(diagramText, context)
```
Mogztter, line 48: Just a nitpick but I believe it will be easier to migrate to ESM if we have explicit extension:
```suggestion
    diagramText = require('./preprocess.js').preprocessPlantUML(diagramText, context)
```

(and it's consistent with the require just above)

src/preprocess.js
```diff
@@ -56,6 +58,210 @@ ${diagramText}
   return JSON.stringify(diagramObject)
 }
 
+// @ts-check
```
Mogztter, line 61: nitpick: `// @ts-check` is already present line 3, do we need this comment here?

src/preprocess.js
```diff
@@ -56,6 +58,210 @@ ${diagramText}
...
+ * @returns {string}
+ */
+module.exports.preprocessPlantUML = function (diagramText, context) {
+  let vfs = context.vfs
+  if (typeof vfs === 'undefined' || typeof vfs.read !== 'function' || typeof vfs.exists !== 'function') {
+    vfs = require('./node-fs.js')
+  }
+  const includeOnce = []
+  const includeStack = []
+  return preprocessPlantUmlIncludes(diagramText, '.', includeOnce, includeStack, vfs)
```
Mogztter, line 74: I'm not sure how relative paths are resolved on PlantUML. Is the relative path resolved relatively to the PlantUML file itself?

```
/path/to/dir/a.puml
/path/to/b.puml
/path/to/c.puml
/path/to/dir/subdir/d.puml
/path/to/e.puml
```

`/path/to/dir/a.puml` includes `../b.puml`
`/path/to/b.puml` includes `./c.puml`
`c.puml` includes `./dir/subdir/d.puml`
`./dir/subdir/d.puml` includes `../../e.puml`

Is that how it's supposed to work?

src/asciidoctor-kroki.js
anb0s: done

src/preprocess.js
anb0s: done

src/preprocess.js
anb0s: Yes, that how it's supposed to work - PlantUML resolves the included files relative to the file that includes them.

Mogztter:
I will probably checkout your branch tomorrow and carefully review the code :eyes:

Mogztter:
@anb0s Overall this is really good and I've made only a few tiny changes. The biggest change is that the preprocessor now preserves comments. I think it's important otherwise we will "lose" information if we want to use the encoded format (ie. the URL) as the "source of trust".
I don't recommend it but some users asked me if all the content was in the encoded format so I think it's important to preserve comments.
I've also added a few tests (but most of the cases were already covered :+1:) and I also resolve the "first" `!include` directive relative to the base directory instead of `.`. For instance:

```
[plantuml]
....
!include ./style-general.iuml
alice -> bob
....
```

If you base directory is `/path/to/docs`, the preprocessor will try to read `/path/to/docs/style-general.iuml`. If base directory is undefined, I believe that the value will be "the current working directory".

Let me know what you think.

anb0s:
> @anb0s Overall this is really good and I've made only a few tiny changes. The biggest change is that the preprocessor now preserves comments. I think it's important otherwise we will "lose" information if we want to use the encoded format (ie. the URL) as the "source of trust".
> I don't recommend it but some users asked me if all the content was in the encoded format so I think it's important to preserve comments.

I've also thought first to preserve them, but then decided as not needed and easier to remove instead of "handle". So i'm fine with preserving comments 👍 

> I've also added a few tests (but most of the cases were already covered 👍) and I also resolve the "first" `!include` directive relative to the base directory instead of `.`.

Yes, that's also good to have it explicitly defined instead of expecting working directory be the same as base directory of the document.

> If you base directory is `/path/to/docs`, the preprocessor will try to read `/path/to/docs/style-general.iuml`. If base directory is undefined, I believe that the value will be "the current working directory".

Yes it looks like it's the current working directory if not defined.

> Let me know what you think.

I've reviewed the changes and it looks good to me. I can maybe try it tomorrow, but if all tests are green i'm fine with this 💯 

Thank You!

anb0s:
I've checked out the new version and it looks good to me!

From my point of view it can be merged ;)

Mogztter:
Let's do this!
A big thanks for your work Andre 👍

PR #77, opened 2020-05-17
resolves #42 use a POST request if the URI is too large

### LINKED ISSUE #42: Switch to POST if the URI is too long

Mogztter:
resolves #42


Mogztter:
Damn!
```
Error: POST https://kroki.io/vegalite/svg - error; reason: spawnSync C:\windows\system32\cmd.exe ENAMETOOLONG
```
I guess `unxhr` will need to write a file to disk to pass the data to the request... :disappointed:

Mogztter:
Woot green bars :tada:

Issue #90, opened 2020-06-04
Find a way to not fetch the same image again when using a user-defined filename


Mogztter:
The filename of the image can be defined by the user. In this case, we cannot tell if we need to fetch the image again because the content has changed or if we don't because the content is the same.

Previously we were using the md5sum of the diagram content as filename.
We might need to write a metadata file next to the images to compare the md5sum.

Mogztter:
We should leverage XDG https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html and create a single JSON file that contains a map of filename with the associated md5sum of the source diagram.

//cc @danyill

danyill:
How about post-processing svg data to add the metadata in there rather than having a manifest? 

The same thing could be done with png (see the [spec](https://www.w3.org/TR/PNG/#11textinfo)). That would be at some level "nicer" than publishing a manifest.

Do we support other asciidoctor-diagram formats? (pdf and txt)?

It would probably add somewhat to processing time but might be a "cleaner" although more intricate solution (and perhaps harder to support for the browser).

Thoughts?

Mogztter:
> Do we support other asciidoctor-diagram formats? (pdf and txt)?

Yes we support txt, pdf and jpeg but I think we can safely ignore pdf in this context.

> It would probably add somewhat to processing time but might be a "cleaner" although more intricate solution (and perhaps harder to support for the browser).

Indeed, it means that we need libraries to understand png, svg and jpeg formats to add the additional metadata.
In a browser, we do not fetch the images so that won't be an issue until we do.

> The same thing could be done with png (see the spec). That would be at some level "nicer" than publishing a manifest.

It might be one of those cases where "le mieux est l'ennemi du bien." (not sure how to translate that, probably best/perfect is the enemy of the good).
In other words, it sounds fancy and nice but I think we should keep it simple. I'm reluctant to alter the content of the images but it might be something we could implement in the Kroki server...?

danyill:
Thanks for your thoughts @Mogztter I just wanted to add this idea into the mix. Yes we have the same expression "the perfect is the enemy of the good".

Mogztter:
> Maybe we should revert 00a8a69 and release a new version.

I will move forward and do that. It will give us some time to think it through.

> Thanks for your thoughts @Mogztter I just wanted to add this idea into the mix. Yes we have the same expression "the perfect is the enemy of the good".

[...]

Issue #48, opened 2020-04-20
[kroki-fetch-diagram] generated-file-name parameter not used


anb0s:
see #47 for my setup

My test presentation.adoc sets the `imagesdir` and use `generated-file-name` [parameter ](https://github.com/asciidoctor/asciidoctor-diagram#creating-a-diagram)to have all images with human readable names:
```
ifndef::imagesdir[:imagesdir: ./images]

== UML class diagram test with kroki
[plantuml, "diagram-classes", svg]
....
class BlockProcessor
class DiagramBlock
class DitaaBlock
class PlantUmlBlock

BlockProcessor <|-- DiagramBlock
DiagramBlock <|-- DitaaBlock
DiagramBlock <|-- PlantUmlBlock
....
```

It fetches the diagram and writes to `images/210a0ebf2a7bb40669c4e3c895b2c9396f048327.svg` instead of `images/diagram-classes.svg` like ascidoctor-diagram does.

It would be good to support the mandatory diagram parameters from asciidoctor-diagram.

Mogztter:
> It fetches the diagram and writes to images/210a0ebf2a7bb40669c4e3c895b2c9396f048327.svg instead of images/diagram-classes.svg like ascidoctor-diagram does.

I need to check how Asciidoctor diagram resolve conflicts:

```adoc
[plantuml, "diagram-classes", svg]
....
....

[plantuml, "diagram-classes", svg]
....
....
```

anb0s:
Yes it overwrites the file and it's OK, because i think conflict resolution would be complex to implement and user is responsible to use unique names or just not enter any generated-file-name to have it generated and unique...

Mogztter:
@anb0s If you want to give it a try, the line to update is:

https://github.com/Mogztter/asciidoctor-kroki/blob/b0575d075ea7088695f14318ca2969091e5b0267/src/fetch.js#L20

The "diagramName" (ie. file name) should be: `${target}.${format}`.
Once it's done, we should probably remove the dependency on `rusha` since we won't need it anymore and update and/or add a test to make sure that it's working as expected.

anb0s:
> @anb0s If you want to give it a try, the line to update is:
> 
> https://github.com/Mogztter/asciidoctor-kroki/blob/b0575d075ea7088695f14318ca2969091e5b0267/src/fetch.js#L20
> 
> The "diagramName" (ie. file name) should be: `${target}.${format}`.

[...]

PR #94, opened 2020-06-13
Revert "Use second optional positional attribute (target) as generated-file-name"

Mogztter:
This reverts commit 00a8a69d32bb19939e151126802e04397ad6fd60.


anb0s:
I'm back again - we need this feature in the released version :(

anb0s:
OK, i see now the issue #90 and reopened #48

Mogztter:
Yes we are using a very basic cache system based on the filename but if the user can control the filename it won't work so we need to find a solution.

PR #96, opened 2020-06-15
resolves #92 move Ruby extension into ruby subdirectory

### LINKED ISSUE #92: Move the code of the Ruby extension in a subproject
Currently, the code is available at: https://github.com/asciidoctor/asciidoctor-intellij-plugin/blob/master/src/main/resources/kroki-extension.rb

We should create a proper Ruby project with tests and publish the gem to https://rubygems.org/

Mogztter:
resolves #92


ruby/lib/asciidoctor/extensions/asciidoctor_kroki/extension.rb
```diff
@@ -0,0 +1,121 @@
...
+          end
+        else
+          'kroki'
+        end
+      end
+
+      def get_format(attrs, diagram_type)
+        format = attrs['format'] || 'svg'
+        # The JavaFX preview doesn't support SVG well, therefore we'll use PNG format...
+        if format == 'svg'
```
Mogztter, line 102: We should check if `env-idea` attribute is defined.

Issue #72, opened 2020-05-12
Document the macro form


danyill:
Currently we don't document the macro form for use with asciidoctor-kroki and initially I thought it wasn't supported based on the README. 

Looking through the tests I joyfully realise it is indeed possible!

```asciidoc
plantuml::${file}[svg,role=sequence]
vegalite::test/fixtures/chart.vlite[svg,role=chart]
bytefield::test/fixtures/simple.bytefield[svg,role=bytefield,opts=inline]
vegalite::test/fixtures/chart.vlite[svg,role=chart,opts=interactive]
```

I'll do a PR for this at some stage.

Mogztter:
https://blog.codinghorror.com/if-it-isnt-documented-it-doesnt-exist/ :sweat_smile: 

> I'll do a PR for this at some stage.

That would be great :grinning:

juracy:
I think the macro form is already in README.md

By the way, why not README.adoc ?

Mogztter:
> I think the macro form is already in README.md

You're right, I will close this issue.

> By the way, why not README.adoc 

The main reason is that npmjs only supports Markdown: https://www.npmjs.com/package/asciidoctor-kroki
It's sad but what can we do... 😞

PR #100, opened 2020-06-20
resolves #39 document attributes and options

### LINKED ISSUE #39: Document configuration options
Bits and pieces I learned in discussions that should be documented:

**General diagram options**
> here: https://github.com/asciidoctor/asciidoctor-diagram
> we are trying to stay compatible with Asciidoctor diagram

> please note that diagram attributes are not supported since the Kroki API does not support specific attributes: https://github.com/asciidoctor/asciidoctor-diagram#diagram-attributes

**All Asciidoctor Kroki parameters**

Besides `kroki-server-url` and `kroki-fetch-diagram` there are probably others?

Mogztter:
As always, review is highly appreciated :smile: 

Ping @eshepelyuk @danyill @marcelstoer 

resolves #39

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,109 @@ include::alice-bob.puml[]
 
 ![PlantUML diagram](https://kroki.io/plantuml/png/eNpzKC5JLCopzc3hSszJTE5V0LVTSMpP4nJIzUsBCgIApPUKcg==)
 
+### Syntax
+
+You can declare positional and named attributes when using the block or the macro form.
+
+**Positional attributes**
+
+When using the block form:
+
+1. The first positional attribute specifies the diagram type (see below for a complete list).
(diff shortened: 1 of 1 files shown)


Mogztter:
Related issue #72
I mention both the block and macro forms. Maybe we should add a dedicated section before to explain the two forms? What do you think?

README.md
```diff
@@ -175,6 +175,109 @@ include::alice-bob.puml[]
...
+....
+graph TD;
+    A-->B;
+    A-->C;
+    B-->D;
+    C-->D;
+....
+```
+
+In the above example,
```
marcelstoer, line 203: This renders as a single line despite the line breaks here (standard Markdown behavior). If that's unintended you need to add line breaks or define a list.

marcelstoer:
That's fantastic! Thank you.

> Maybe we should add a dedicated section before to explain the two forms? What do you think?

Either that or link to explanations elsewhere.

README.md
Mogztter: Yes this is intended, I find it easier to edit.

PR #102, opened 2020-06-27
Set path when updating the Antora content catalog

bwalding:
People directly using the Antora content catalog (it's a long story :) ) may need the path set for consistency with the rest of the entries in there.

I'm not sure if it makes more sense to pass the full path around with the image, but in the interest of changing the least code, I've just added the path when setting the file in the content catalog.

### DIFF
--- a/src/antora-adapter.js
+++ b/src/antora-adapter.js
@@ -17,6 +17,7 @@ module.exports = (file, contentCatalog, vfs) => {
             module,
             family: 'image',
             mediaType: image.mediaType,
+            path: image.relative + '/' + image.basename,
             basename: image.basename,
             relative: image.basename
           }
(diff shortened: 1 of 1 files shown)


Mogztter:
Hello @bwalding 

I think it's a reasonable change

src/antora-adapter.js
```diff
@@ -17,6 +17,7 @@ module.exports = (file, contentCatalog, vfs) => {
             module,
             family: 'image',
             mediaType: image.mediaType,
+            path: image.relative + '/' + image.basename,
```
Mogztter, line 20: We should probably use the module `path` to join the path:
```suggestion
            path: path.join(image.relative, image.basename),
```

Mogztter:
LGTM, thanks Ben :smile:

Issue #104, opened 2020-07-01
0.9.1 is not compatible with Antora, which is still stuck on asciidoctor 1.5.8 < 2


djencks:
Unfortunately some asciidoctor 2-isms crept into the latest 0.9.1 release, making it unusable with Antora.

PR #106, opened 2020-07-01
resolves #87: implement kroki-default-format and kroki-default-options attributes

### LINKED ISSUE #87: Support global attributes to configure all diagrams
We should probably configure the (default) output format.

djencks:
This makes asciidoctor-kroki respond to kroki-default-format and kroki-default-options attributes.

kroki-default-format can be svg or png.  The purpose of allowing svg here is to allow resetting to default in a clearer way than unsetting the attribute.

kroki-default-options can be inline, interactive, or none.
In the block or block macro, `opts=none` will undo the effect of setting kroki-default-options.

This is implemented on top of #104.


README.md
```diff
@@ -126,6 +126,7 @@ If you are using [Antora](https://antora.org/), you can integrate Kroki in your
 **💡 TIP**:
 You can use the `kroki-fetch-diagram` option to download the images from Kroki at build time.
 In other words, while viewing pages you won't rely on Kroki anymore.
+However, this is not compatible with inline or interactive svg images.
```
Mogztter, line 129: I'm not 100% sure but I think it's working with interactive SVG images.

README.md
```diff
@@ -314,6 +316,27 @@ Consult the [Kroki documentation](https://kroki.io/#support) to find out which f
 | `kroki-fetch-diagram` | Define if we should download (and save on the disk) the images from the Kroki server.<br/>This feature is not available when running in the browser. | `false`
 | `kroki-http-method` | Define how we should get the image from the Kroki server. Possible values:<br/><ul><li>`get`: always use GET requests</li><li>`post`: always use POST requests</li><li>`adaptive`: use a POST request if the URI length is longer than 4096 characters, otherwise use a GET request</li></ul> | `adaptive` |
 
+### Default configuration
+
+By default, images are generated as svg when possible.
+To alter this, set `kroki-default-format` as a global, header, or page attribute:
```
Mogztter, line 322: It might be misleading, attributes can be assigned to a document:
- through the CLI or API
- in the document’s header
- in the document’s body

A "page attribute" is definitely an Antora concept, you probably meant document body attribute?

README.md
```diff
@@ -314,6 +316,27 @@ Consult the [Kroki documentation](https://kroki.io/#support) to find out which f
...
+
+```adoc
+:kroki-default-format: png
+```
+
+You can unset this with `:kroki-default-format!:` or `:kroki-default-format: svg`.
+
+By default, asciidoctorkroki generates a link, to a kroki server or a local file.
```
Mogztter, line 330:
```suggestion
By default, Asciidoctor Kroki generates a link, to a Kroki server or a local file.
```

djencks: Agreed.

README.md
djencks: I have a different wording now, WDYT?

src/asciidoctor-kroki.js
```diff
@@ -67,13 +71,16 @@ const processKroki = (processor, parent, attrs, diagramType, diagramText, contex
...
-  const interactive = attrs['interactive-option'] === ''
-  if (interactive) {
-    blockAttrs['interactive-option'] = ''
+
+  const opt = choices.reduce((accum, [name, option]) => {
+    return attrs[name + '-option'] === '' ? option : accum
+  }, undefined) ||
+      choices.reduce((accum, [name, option]) => {
+        return parent.document.getAttribute('kroki-default-options') === name ? option : accum
+      }, undefined)
```
Mogztter, line 80: Not sure about this code, it might be elegant but I cannot understand it. If we want to make the code more concise I guess we could introduce a function to check a given list of option but if we can avoid using `reduce` that would be great (since it's not consistent with the code style and not straightforward to read in my opinion)

src/asciidoctor-kroki.js
```diff
@@ -48,7 +52,7 @@ const processKroki = (processor, parent, attrs, diagramType, diagramText, contex
     diagramText = require('./preprocess.js').preprocessPlantUML(diagramText, context, doc.getBaseDir())
   }
   const blockId = attrs.id
-  const format = attrs.format || 'svg'
+  const format = attrs.format || parent.document.getAttribute('kroki-default-format') || 'svg'
```
Mogztter, line 55: Asciidoctor (and by extension Asciidoctor.js) provides a function to resolve and attribute:

https://github.com/asciidoctor/asciidoctor.js/blob/2b006467bd99b2ce8589b107e8fc34a6a4323880/packages/core/src/asciidoctor-core-api.js#L903-L906

But I'm fine with the `||`.

Mogztter:
@djencks Thanks! I've merged #104 could you please rebase?

src/asciidoctor-kroki.js
djencks: Indeed, it might be write-only code, like xslt :-) It checks for whether any block/macro attribute option is set, and if none, it checks for the global/default option, and returns the appropriate option name.  I'll see if I can make it more transparent :-)  but it was the simplest code I could actually get to work.

src/asciidoctor-kroki.js
djencks: I think the useful form of that function, with the fallback to the alternate name, is only available in asciidoctor 2.  In order to avoid potential confusion about which version of the function is being used I'd like to stick with the ||.

src/asciidoctor-kroki.js
djencks: See how you like it now...

README.md
```diff
@@ -314,6 +316,28 @@ Consult the [Kroki documentation](https://kroki.io/#support) to find out which f
...
+### Default configuration
+
+By default, images are generated as svg when possible.
+To alter this, set `kroki-default-format` as an 'external', document header, or document body attribute:
+
+```adoc
+:kroki-default-format: png
+```
+
+'External' means of setting an attribute vary depending on your environment, but include CLI, API, and Antora playbook and component descriptor attribute settings.
```
Mogztter, line 328: I think we should add this as a note/tip:

**ℹ️ NOTE:** An AsciiDoc attribute can be defined through the CLI or API, in the document’s header or in the document’s body. In addition, if you are using Antora, you can define AsciiDoc attributes in your playbook and/or in your component descriptor.

References:

- https://asciidoctor.org/docs/user-manual/#setting-attributes-on-a-document
- https://docs.antora.org/antora/2.3/page/attributes/#custom-attributes

I don't want to give the impression that this is something specific to this extension but rather that we are using a built-in feature (an AsciiDoc attribute).
I don't mind "duplicating" the documentation to explain where an attribute can be defined. 

Maybe we should add a concrete example? For instance: "it's possible to define the output format globally in an Antora documentation site if we define 	`:kroki-default-format: png` in the playbook."

package.json
```diff
@@ -12,6 +12,7 @@
     "test:node": "mocha test/**.spec.js",
     "test:browser": "node test/browser/run.js",
     "lint": "standard src/**.js test/**.js",
+    "lint-fix": "standard src/**.js test/**.js --fix",
```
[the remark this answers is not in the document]
djencks: That doesn't work for me.  No error from the command line, but also no fixes are applied.

README.md
djencks: I made pretty much this change.  I don't understand what the ℹ️ character is, I just copied and pasted it, is it reasonably correct?

package.json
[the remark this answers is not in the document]
Mogztter: I meant:

```suggestion
    "lint-fix": "npm run lint -- --fix",
```

So you can run `npm run lint-fix` (arguably easier to type than `npm run lint -- --fix`) but we don't duplicate "standard src/**.js test/**.js".

src/asciidoctor-kroki.js
```diff
@@ -67,13 +71,24 @@ const processKroki = (processor, parent, attrs, diagramType, diagramText, contex
...
+  choices.forEach(([name, option]) => {
+    if (attrs[name + '-option'] === '') {
+      opt = option
+    }
+  })
+  opt || choices.forEach(([name, option]) => {
+    if (parent.document.getAttribute('kroki-default-options') === name) {
+      opt = option
+    }
+  })
```
Mogztter, line 88: I can read it :smile: 

I would suggest:

```js
const availableOptions = ['inline', 'interactive', 'none']
```

```suggestion

  let currentOption
  for (availableOption of availableOptions) {
    if (attrs[`${availableOption}-option`] === '') {
      currentOption = availableOption
    }
  }
  if (!currentOption) {
    for (availableOption of availableOptions) {
      if (parent.document.getAttribute('kroki-default-options') === availableOption) {
        currentOption = availableOption
      }
    }
  }
  if (currentOption && currentOption !== 'none') {
    blockAttrs[`${currentOption}-option`] = ''
  }
```

And probably extract the code in a function:

```js
const currentOption = getCurrentOption(attrs, parent.document)
if (currentOption && currentOption !== 'none') {
  blockAttrs[`${currentOption}-option`] = ''
}
```

README.md
Mogztter: Yes it's an hack to emulate an admonition using Markdown

Mogztter:
The code looks fine, I can take care of the last remaining nitpicks if you want (as they are more subjective)

src/asciidoctor-kroki.js
djencks: A separate function lets the code be a lot simpler!

src/asciidoctor-kroki.js
```diff
@@ -34,6 +36,22 @@ const createImageSrc = (doc, krokiDiagram, target, vfs, krokiClient) => {
   return imageUrl
 }
 
+function getOption (attrs, parent) {
```
Mogztter, line 39: :+1:

djencks:
Feel free to change anything else you can think of and apply this!

Mogztter:
Thanks @djencks for taking care of it, greatly appreciated :+1:

PR #110, opened 2020-07-05
WIP resolves #109 Add kref support

### LINKED ISSUE #109: Consider supporting xref like syntax in diagrams, translated to diagram specific syntax
This may be mostly useful in Antora.
Many types of diagrams, in particular plantuml, support linking from svg diagram elements to elsewhere.
The idea, which Ec Roc thought of, is to allow specifying these links using xref-like syntax, and translating them into the diagram specific syntax.
Since this is kroki I propose calling them 'kref'.
I've implemented a prototype of this as a standalone project, https://gitlab.com/djencks/asciidoctor-antora-kref, but it might make sense to  integrate this into asciidoctor-kroki so the block/block macro can communicate the diagram type to the 'kref' inline macro so the desired format does not need to be further specified.
Since I don't have any idea what interpretation or translation of 'xref' syntax might be appropriate outside Antora, I've started by only registering the kref inline macro when Antora is present.

djencks:
Initial commit has minimal changes from @djencks/asciidoctor-antora-kref and installs the kref inline macro only when Antora is present.
The only additions are communicating the diagramType to the kref inline macro, and indexing formats on diagram type rather than format name.
This is not a finished product, but a basis for discussion and investigation.

### DIFF
--- a/src/asciidoctor-kroki.js
+++ b/src/asciidoctor-kroki.js
@@ -64,7 +64,9 @@ const processKroki = (processor, parent, attrs, diagramType, diagramText, contex
   // Be careful not to specify "specialcharacters" or your diagram code won't be valid anymore!
   const subs = attrs.subs
   if (subs) {
+    context.diagramType = diagramType
     diagramText = parent.applySubstitutions(diagramText, parent.$resolve_subs(subs))
+    delete context.diagramType
   }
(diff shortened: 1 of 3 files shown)


Mogztter:
I think it would be better to open an issue so we can precisely describe the use case and how we can address it.

PR #119, opened 2020-07-21
Support for plantuml-config param

### LINKED ISSUE #118: Support plantuml-config-attribute for site-wide includes
The asciidoctor-plantuml-extension had a very cool option `plantuml-config-file` that allowed default-includes for all diagrams, so that they had the same design.

I'm aware of the plantuml-server include version, but I especially liked that the config could be checked in together with the document and that each document was able to define their own config without needing access to the server.

Could we implement such an options for Kroki?

gtudan:
This adds a new param for a common config for all plantuml diagrams.
The config will be added to the top of the diagram. It allows users to define a common skin.

resolves #118

### DIFF
--- a/README.md
+++ b/README.md
@@ -315,8 +315,7 @@ Consult the [Kroki documentation](https://kroki.io/#support) to find out which f
 | `kroki-server-url` | The URL of the Kroki server (see "Using Your Own Kroki") | `https://kroki.io`
 | `kroki-fetch-diagram` | Define if we should download (and save on the disk) the images from the Kroki server.<br/>This feature is not available when running in the browser. | `false`
 | `kroki-http-method` | Define how we should get the image from the Kroki server. Possible values:<br/><ul><li>`get`: always use GET requests</li><li>`post`: always use POST requests</li><li>`adaptive`: use a POST request if the URI length is longer than 4096 characters, otherwise use a GET request</li></ul> | `adaptive` |
-| `kroki-plantuml-include` | A file that will be included at the top of all PlantUML diagrams as if `!include file` was used. This allows you to define a common skin for your diagrams. This can be a path or a URL. |  |
-
+| `kroki-plantuml-include` | A file that will be included at the top of all PlantUML diagrams as if `!include file` was used. This can be useful when you want to define a common skin for all your diagrams. The value can be a path or a URL. |  |
 ### Default configuration
 
 By default, images are generated as SVG when possible.
--- a/README.md
+++ b/README.md
@@ -315,7 +315,7 @@ Consult the [Kroki documentation](https://kroki.io/#support) to find out which f
 | `kroki-server-url` | The URL of the Kroki server (see "Using Your Own Kroki") | `https://kroki.io`
 | `kroki-fetch-diagram` | Define if we should download (and save on the disk) the images from the Kroki server.<br/>This feature is not available when running in the browser. | `false`
 | `kroki-http-method` | Define how we should get the image from the Kroki server. Possible values:<br/><ul><li>`get`: always use GET requests</li><li>`post`: always use POST requests</li><li>`adaptive`: use a POST request if the URI length is longer than 4096 characters, otherwise use a GET request</li></ul> | `adaptive` |
-| `kroki-plantuml-config` | A file that will be included at the top of all PlantUML diagrams. This allows you to define a common skin for your diagrams. This can be a path or a URL. |  |
+| `kroki-plantuml-include` | A file that will be included at the top of all PlantUML diagrams as if `!include file` was used. This allows you to define a common skin for your diagrams. This can be a path or a URL. |  |
 
 ### Default configuration
 
--- a/README.md
+++ b/README.md
@@ -315,6 +315,7 @@ Consult the [Kroki documentation](https://kroki.io/#support) to find out which f
 | `kroki-server-url` | The URL of the Kroki server (see "Using Your Own Kroki") | `https://kroki.io`
 | `kroki-fetch-diagram` | Define if we should download (and save on the disk) the images from the Kroki server.<br/>This feature is not available when running in the browser. | `false`
 | `kroki-http-method` | Define how we should get the image from the Kroki server. Possible values:<br/><ul><li>`get`: always use GET requests</li><li>`post`: always use POST requests</li><li>`adaptive`: use a POST request if the URI length is longer than 4096 characters, otherwise use a GET request</li></ul> | `adaptive` |
+| `kroki-plantuml-config` | A file that will be included at the top of all PlantUML diagrams. This allows you to define a common skin for your diagrams. This can be a path or a URL. |  |
 
 ### Default configuration
 
--- a/ruby/lib/asciidoctor/extensions/asciidoctor_kroki/extension.rb
+++ b/ruby/lib/asciidoctor/extensions/asciidoctor_kroki/extension.rb
@@ -76,8 +76,8 @@ module AsciidoctorExtensions
       private
 
       def prepend_plantuml_config(diagram_text, diagram_type, doc)
-        if diagram_type == :plantuml && doc.attributes['kroki-plantuml-config']
-          config = File.read(doc.attributes['kroki-plantuml-config'])
+        if diagram_type == :plantuml && doc.attributes['kroki-plantuml-include']
+          config = File.read(doc.attributes['kroki-plantuml-include'])
           diagram_text = config + '\n' + diagram_text
         end
         diagram_text
--- a/ruby/lib/asciidoctor/extensions/asciidoctor_kroki/extension.rb
+++ b/ruby/lib/asciidoctor/extensions/asciidoctor_kroki/extension.rb
@@ -52,6 +52,7 @@ module AsciidoctorExtensions
     class << self
       def process(processor, parent, attrs, diagram_type, diagram_text)
         doc = parent.document
+        diagram_text = prepend_plantuml_config(diagram_text, diagram_type, doc)
         # If "subs" attribute is specified, substitute accordingly.
         # Be careful not to specify "specialcharacters" or your diagram code won't be valid anymore!
         if (subs = attrs['subs'])
@@ -73,6 +74,14 @@ module AsciidoctorExtensions
       end
 
       private
+
+      def prepend_plantuml_config(diagram_text, diagram_type, doc)
+        if diagram_type == :plantuml && doc.attributes['kroki-plantuml-config']
+          config = File.read(doc.attributes['kroki-plantuml-config'])
+          diagram_text = config + '\n' + diagram_text
+        end
+        diagram_text
+      end
 
       def get_alt(attrs)
         if (title = attrs['title'])
--- a/ruby/lib/asciidoctor/extensions/asciidoctor_kroki/extension.rb
+++ b/ruby/lib/asciidoctor/extensions/asciidoctor_kroki/extension.rb
@@ -76,8 +76,10 @@ module AsciidoctorExtensions
       private
 
       def prepend_plantuml_config(diagram_text, diagram_type, doc)
-        if diagram_type == :plantuml && doc.attributes['kroki-plantuml-include']
-          config = File.read(doc.attributes['kroki-plantuml-include'])
+        if diagram_type == :plantuml && doc.attr?('kroki-plantuml-include')
+          # TODO: this behaves different than the JS version
+          # The file should be added by !include #{plantuml_include}" once we have a preprocessor for ruby
+          config = File.read(doc.attr('kroki-plantuml-include'))
           diagram_text = config + '\n' + diagram_text
         end
         diagram_text
--- a/ruby/spec/asciidoctor_kroki_spec.rb
+++ b/ruby/spec/asciidoctor_kroki_spec.rb
@@ -33,5 +33,19 @@ describe ::AsciidoctorExtensions::KrokiBlockProcessor do
 </div>
 </div>)
     end
+    it 'should include the plantuml-config file' do
+      input = <<~'ADOC'
+        [plantuml]
+        ....
+        alice -> bob: hello
+        ....
+      ADOC
+      output = Asciidoctor.convert(input, attributes: { 'env-idea' => '', 'kroki-plantuml-config' => 'spec/fixtures/config.puml' }, standalone: false)
+      (expect output).to eql %(<div class="imageblock kroki">
+<div class="content">
+<img src="https://kroki.io/plantuml/png/eNorzs7MK0gsSsxVyM3Py0_OKMrPTVUoKSpN5YrJS8zJTE5V0LVTSMpPslLISM3JyQcArVsRHA==" alt="Diagram">
+</div>
+</div>)
+    end
   end
 end
--- a/ruby/spec/asciidoctor_kroki_spec.rb
+++ b/ruby/spec/asciidoctor_kroki_spec.rb
@@ -33,14 +33,14 @@ describe ::AsciidoctorExtensions::KrokiBlockProcessor do
 </div>
 </div>)
     end
-    it 'should include the plantuml-config file' do
+    it 'should include the plantuml-include file' do
       input = <<~'ADOC'
         [plantuml]
         ....
         alice -> bob: hello
         ....
       ADOC
-      output = Asciidoctor.convert(input, attributes: { 'env-idea' => '', 'kroki-plantuml-config' => 'spec/fixtures/config.puml' }, standalone: false)
+      output = Asciidoctor.convert(input, attributes: { 'env-idea' => '', 'kroki-plantuml-include' => 'spec/fixtures/config.puml' }, standalone: false)
       (expect output).to eql %(<div class="imageblock kroki">
 <div class="content">
 <img src="https://kroki.io/plantuml/png/eNorzs7MK0gsSsxVyM3Py0_OKMrPTVUoKSpN5YrJS8zJTE5V0LVTSMpPslLISM3JyQcArVsRHA==" alt="Diagram">
--- a/ruby/spec/fixtures/config.puml
+++ b/ruby/spec/fixtures/config.puml
@@ -1,0 +1,1 @@
+skinparam monochrome true
--- a/src/asciidoctor-kroki.js
+++ b/src/asciidoctor-kroki.js
@@ -69,9 +69,9 @@ const processKroki = (processor, parent, attrs, diagramType, diagramText, contex
   if (diagramType === 'vegalite') {
     diagramText = require('./preprocess.js').preprocessVegaLite(diagramText, context)
   } else if (diagramType === 'plantuml') {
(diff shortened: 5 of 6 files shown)


Mogztter:
Thanks for your contribution!
Tests failures are probably unrelated, see: https://github.com/Mogztter/asciidoctor-kroki/pull/117#issuecomment-661207566
I will try to fix them in the next few days.

gtudan:
Yeah, they fail on master as well.
I also added the parameter in the ruby version

Mogztter:
> I also added the parameter in the ruby version

Awesome, thanks 👍 

> Yeah, they fail on master as well.

Could you please rebase on master, the tests are now running fine on master 😅

gtudan:
Done - looks good to me

Mogztter:
@gtudan Might be a new line issue on Windows, you can use `encode` or `encodeText` to make sure that the expected value will use the same new line characters (`\n` on Linux and `\n\r` on Windows):

https://github.com/Mogztter/asciidoctor-kroki/blob/f8b706afca3b49caf74a6f044e0217ef826f5d74/test/test.spec.js#L40-L52

gtudan:
You're right about the line endings:

```
$ diff -u real expected | cat -vet
--- real^I2020-07-23 16:24:26.000000000 +0200$
+++ expected^I2020-07-23 16:24:36.000000000 +0200$
@@ -1,5 +1,5 @@$
-skinparam DefaultFontName "Neucha"$
+skinparam DefaultFontName "Neucha"^M$
 skinparam BackgroundColor black$
-@startuml$
-alice -> bob$
+@startuml^M$
+alice -> bob^M$
 @enduml$
```

I don't think encoding the text helps here, because the line endings appear inside the diagram, and not at the end.
I worked around this by decoding the diagram from the url and asserting the individual parts.

There is a minor problem here: if the diagram starts with `@startuml` then the skin will end up before the preamble. asciidoctor-plantuml also does it this way, but I never noticed the problem there. Maybe it's because I seldom use the start/end-tags?

Mogztter:
I've just noticed that the `-config` option has been removed from PlantUML?

```console
$ java -jar plantuml.jar -help          
Usage: java -jar plantuml.jar [options] -gui
	(to execute the GUI)
    or java -jar plantuml.jar [options] [file/dir] [file/dir] [file/dir]
	(to process files or directories)

You can use the following wildcards in files/dirs:
	*	means any characters but '/'
	?	one and only one character but '/'
	**	means any characters (used to recurse through directories)

where options include:
    -gui		To run the graphical user interface
    -tpng		To generate images using PNG format (default)
    -tsvg		To generate images using SVG format
    -teps		To generate images using EPS format
    -tpdf		To generate images using PDF format
    -tvdx		To generate images using VDX format
    -txmi		To generate XMI file for class diagram
    -tscxml		To generate SCXML file for state diagram
    -thtml		To generate HTML file for class diagram
    -ttxt		To generate images with ASCII art
    -tutxt		To generate images with ASCII art using Unicode characters
    -tlatex		To generate images using LaTeX/Tikz format
    -tlatex:nopreamble	To generate images using LaTeX/Tikz format without preamble
    -o[utput] "dir"	To generate images in the specified directory
    -DVAR1=value	To set a preprocessing variable as if '!define VAR1 value' were used
    -Sparam1=value	To set a skin parameter as if 'skinparam param1 value' were used
    -r[ecurse]		recurse through directories
    -I/path/to/file	To include file as if '!include file' were used
    -I/path/to/*.puml	To include files with pattern
    -charset xxx	To use a specific charset (default is UTF-8)
    -e[x]clude pattern	To exclude files that match the provided pattern
    -metadata		To retrieve PlantUML sources from PNG images
    -nometadata		To NOT export metadata in PNG/SVG generated files
    -checkmetadata		Skip PNG files that don't need to be regenerated
    -version		To display information about PlantUML and Java versions
    -v[erbose]		To have log information
    -quiet		To NOT print error message into the console
    -debugsvek		To generate intermediate svek files
    -h[elp]		To display this help message
    -testdot		To test the installation of graphviz
    -graphvizdot "exe"	To specify dot executable
    -p[ipe]		To use stdin for PlantUML source and stdout for PNG/SVG/EPS generation
    -encodesprite 4|8|16[z] "file"	To encode a sprite at gray level (z for compression) from an image
    -computeurl|-encodeurl	To compute the encoded URL of a PlantUML source file
    -decodeurl		To retrieve the PlantUML source from an encoded URL
    -syntax		To report any syntax error from standard input without generating images
    -language		To print the list of PlantUML keywords
    -checkonly		To check the syntax of files without generating images
    -failfast		To stop processing as soon as a syntax error in diagram occurs
    -failfast2		To do a first syntax check before processing files, to fail even faster
    -pattern		To print the list of Regular Expression used by PlantUML
    -duration		To print the duration of complete diagrams processing
    -nbthread N		To use (N) threads for processing
    -nbthread auto	To use 8 threads for processing
    -timeout N		Processing timeout in (N) seconds. Defaults to 15 minutes (900 seconds).
    -author[s]		To print information about PlantUML authors
    -overwrite		To allow to overwrite read only files
    -printfonts		To print fonts available on your system
    -enablestats	To enable statistics computation
    -disablestats	To disable statistics computation (default)
    -htmlstats		To output general statistics in file plantuml-stats.html
    -xmlstats		To output general statistics in file plantuml-stats.xml
    -realtimestats	To generate statistics on the fly rather than at the end
    -loopstats		To continuously print statistics about usage
    -splash		To display a splash screen with some progress bar
    -progress		To display a textual progress bar in console
    -pipeimageindex N	To generate the Nth image with pipe option
    -stdlib		To print standard library info
    -extractstdlib	To extract PlantUML Standard Library into stdlib folder
    -filename "example.puml"	To override %filename% variable
    -preproc		To output preprocessor text of diagrams
    -cypher		To cypher texts of diagrams so that you can share them

If needed, you can setup the environment variable GRAPHVIZ_DOT.
```

But it's possible to include files with `-I`, so we might rename the parameter to align. It would also describe explicitly the behavior. 
Argubaly, `plantuml-config` can be confusing...

anb0s:
### About the name
The command `-config "file"` is still wroking with PlantUML Version 1.2020.15.
The older versions (before Preprocessor v2) showed this help:

```
-I/path/to/file                Include file as if '!include file' was used
-I/path/to/*.puml              Include files with pattern
...
-config "file"                 Include a file before each diagram (same as -I ??)
```

I've tried `-I/path/to/file` and it works same as `-config "file"`. I think if we want this feature we should use another clearer name like `kroki-plantuml-include` or `kroki-plantuml-pre-include` or `kroki-plantuml-include-always`

### About the include before @startuml
The plantuml.jar includes it _the right way_ and it works with start/end-tags. Therefore if we want this feature working as expected we have to pass this parameter to preprocessPlantUML and handle it there.
- if pre-include parameter is valid search for `@startuml` tag / line:
  - if found add `!include pre-include` after the line
  - else add `!include pre-include` before diagram-text
- preprocess diagram-text

Mogztter:
Thanks for your input @anb0s 

> The command -config "file" is still wroking with PlantUML Version 1.2020.15.
The older versions (before Preprocessor v2) showed this help...

They are very conservative because it seems that this option has been removed from the help message 3 years ago: https://github.com/plantuml/plantuml/blame/9ba7d08d5247cac5ec47a5f0052da3aaad0e8479/src/net/sourceforge/plantuml/OptionPrint.java#L103

I didn't find an explicit commit message that explains the reason but I'm pretty sure it was deprecated/removed in favor of `-I`.

> I've tried -I/path/to/file and it works same as -config "file". I think if we want this feature we should use another clearer name like kroki-plantuml-include or kroki-plantuml-pre-include or kroki-plantuml-include-always

I think `kroki-plantuml-include` is fine (especially if we want to support including files with pattern).

> About the include before @startuml

I think it's reasonable to handle it in the 
Since we are already preprocessing the content, I think it's reasonable to re-implement the logic in the `preprocessPlantUML` function but we need to carefully read PlantUML source code.

For instance, where the include goes in the following example:

```
@startuml(id=MY_OWN_ID1)
A -> A : stuff1
B -> B : stuff2
@enduml

@startuml(id=MY_OWN_ID2)
C -> C : stuff3
D -> D : stuff4
@enduml
```

gtudan:
Wow, that escalated quickly :-D 
I renamed the parameter from "config" to "include" and updated the description.

Thinking about the main reason why I started this (skinning diagrams) the config/include should end up in both diagrams and that's what plantuml does:

```
$ plantuml -Itest/fixtures/plantuml/style-sequence.iuml -preproc -v two-alices.puml

$ cat  two-alices.preproc                                                          
@startuml(id=MY_OWN_ID1)
skinparam Sequence {
  TitleFontSize               12
  TitleFontColor              #606060
  ArrowColor                  #303030
  DividerBackgroundColor      #EEEEEE
  GroupBackgroundColor        #EEEEEE
  LifeLineBackgroundColor     white
  LifeLineBorderColor         #303030
  ParticipantBackgroundColor  #FEFEFE
  ParticipantBorderColor      #303030
  BoxLineColor                #303030
  BoxBackgroundColor          #DDDDDD
}
A -> A : stuff1
B -> B : stuff2
@enduml

$ cat  two-alices_001.preproc                                                     
@startuml(id=MY_OWN_ID2)
skinparam Sequence {
  TitleFontSize               12
  TitleFontColor              #606060
  ArrowColor                  #303030
  DividerBackgroundColor      #EEEEEE
  GroupBackgroundColor        #EEEEEE
  LifeLineBackgroundColor     white
  LifeLineBorderColor         #303030
  ParticipantBackgroundColor  #FEFEFE
  ParticipantBorderColor      #303030
  BoxLineColor                #303030
  BoxBackgroundColor          #DDDDDD
}
C -> C : stuff3
D -> D : stuff4
@enduml
```

gtudan:
I've updated the code to respect the `@startuml` tag and add the file afterwards.

There are quiet a few corner-cases that are not covered here. I.E. multiple diagrams in one file won't work. Diagrams with ids (`@startuml(id=MY_OWN_ID1)`) neither.

These aren't covered by the current preprocessor yet. I think it would be better to file separate issues for those if that's something we want to support or this PR will get to big.

Mogztter:
> Wow, that escalated quickly :-D

Yep 😆 

> I renamed the parameter from "config" to "include" and updated the description.

Thanks @gtudan!

> Thinking about the main reason why I started this (skinning diagrams) the config/include should end up in both diagrams and that's what plantuml does:

Not sure because I've noticed that when using the following definition:

```
@startuml(id=MY_OWN_ID1)
A -> A : stuff1
B -> B : stuff2
@enduml
                                               
@startuml(id=MY_OWN_ID2)
C -> C : stuff3
D -> D : stuff4
@enduml
```

The PlantUML server will actually render:

```
@startuml
A -> A : stuff1
B -> B : stuff2                                      
C -> C : stuff3
D -> D : stuff4
@enduml
```

So I don't think that we need to append the "kroki-plantuml-include" file in every `@startuml`/`@enduml` block (or even after the first `@startuml`).
The reason is that `@startuml`/`@enduml` tags are optional and, in the end, only one (aggregated) diagram will be rendered.
By optional, I mean that the Kroki server will add the `@startuml` and `@enduml` tag if missing.

So I think we should always prepend the include directive at the top of the file and remove any `@startuml`/`@enduml` directives at the end of the preprocessing.

A few examples, where `:kroki-plantuml-include: style.puml`:

**single without startuml/enduml**

Input:
```
A -> A : stuff1
B -> B : stuff2 
```

Output:
```
!include style.puml
A -> A : stuff1
B -> B : stuff2
```

**single with startuml/enduml**
Input:
```
@startuml
A -> A : stuff1
B -> B : stuff2
@enduml
```

Output:
```
!include style.puml
A -> A : stuff1
B -> B : stuff2
```

**multi with startuml/enduml**

Input:
```
@startuml
A -> A : stuff1
B -> B : stuff2
@enduml

@startuml
C -> C : stuff3
D -> D : stuff4
@enduml
```

Output:
```
!include style.puml
A -> A : stuff1
B -> B : stuff2

C -> C : stuff3
D -> D : stuff4
```

**multi with/without startuml/enduml**

Input:
```
A -> A : stuff1
B -> B : stuff2

@startuml
C -> C : stuff3
D -> D : stuff4
@enduml
```

Output:
```
!include style.puml
A -> A : stuff1
B -> B : stuff2

C -> C : stuff3
D -> D : stuff4
```

@anb0s @gtudan What do you think?

Arguably, the Kroki server could take care of removing the `@startuml`/`@enduml` directives in the "body" of the diagram definition.
In other words, if we send the following definition to Kroki:

```
!include style.puml
A -> A : stuff1
B -> B : stuff2

@startuml
C -> C : stuff3
D -> D : stuff4
@enduml
```

It will be transformed to:

```
@startuml
!include style.puml
A -> A : stuff1
B -> B : stuff2

C -> C : stuff3
D -> D : stuff4
@enduml
```

gtudan:
Does PlantUML behave different in the server version than in cli mode?

Given this diagram:

```
skinparam monochrome true

@startuml
alice -> bob
@enduml
```

The CLI-Version gives me a colored diagram. PlantText.com renders it monochrome.

If I pass a file with two diagrams (like the one above) I get two files in cli mode. The web version seems to ignore the `@startuml` and just renders them as one diagram.

If the server versions ignores `@startuml` tags, we might skip any handling for them and prepend the includes at the top.

anb0s:
> Does PlantUML behave different in the server version than in cli mode?

Oh damn, i've only tested with CLI mode localy this time, but did not noticed the different behaviour with server, because we are always using one "master" puml file with exactly one @startuml/@enduml block that includes others files lyke stile etc.

> If the server versions ignores `@startuml` tags, we might skip any handling for them and prepend the includes at the top.

@gtudan sorry for the extra work, now i'm also for your first solution without any special handling in the preprocessor and just prepend the include

@gtudan @Mogztter should we remove all the @startuml/@enduml tags as last step in the preprocessor, because we do not know (or it may change in future) if the server is handling this for us and calls CLI or may be forward this to roginal PlantUML server?

Mogztter:
> The CLI-Version gives me a colored diagram. PlantText.com renders it monochrome.
If I pass a file with two diagrams (like the one above) I get two files in cli mode. The web version seems to ignore the @startuml and just renders them as one diagram.
If the server versions ignores @startuml tags, we might skip any handling for them and prepend the includes at the top.

I think it makes sense to render only one diagram when using the PlantUML server but in your case I'm not entirely sure if this is intended or not...
In fact, with the following definition, you have a single `@startuml`/`@enduml` block, so the server version of PlantUML should/could render a colored diagram:

```
skinparam monochrome true

@startuml
alice -> bob
@enduml
```

But anyway, that's how the PlantUML server is currently working and we should reproduce this behavior.

> Sorry for the extra work, now i'm also for your first solution without any special handling in the preprocessor and just prepend the include

👍 

> should we remove all the @startuml/@enduml tags as last step in the preprocessor, because we do not know (or it may change in future) if the server is handling this for us and calls CLI or may be forward this to original PlantUML server?

Yes I think we should remove all the @startuml/@enduml tags as last step in the preprocessor. It seems like the easiest solution to have a fully working implementation.

gtudan:
Okay, I reverted my last change. The diagram text will be prepended on top of the diagram.

If you don't mind I'd like to do the stripping of start/enduml tags in a separate MR, as it's going to be quiet a lot of changes (all tests have to be updated since most of them include tags in the expected result)

Mogztter:
> Okay, I reverted my last change. The diagram text will be prepended on top of the diagram.

Thanks :+1:

> If you don't mind I'd like to do the stripping of start/enduml tags in a separate MR, as it's going to be quiet a lot of changes (all tests have to be updated since most of them include tags in the expected result)

That's totally fine. In fact, this new pull request should be merged first.
Let me know if you need help.

For the record, I will be on vacation in August 🗻😃

gtudan:
This really is a tough one - here's my shot at it #124

README.md
```diff
@@ -315,6 +315,7 @@ Consult the [Kroki documentation](https://kroki.io/#support) to find out which f
 | `kroki-server-url` | The URL of the Kroki server (see "Using Your Own Kroki") | `https://kroki.io`
 | `kroki-fetch-diagram` | Define if we should download (and save on the disk) the images from the Kroki server.<br/>This feature is not available when running in the browser. | `false`
 | `kroki-http-method` | Define how we should get the image from the Kroki server. Possible values:<br/><ul><li>`get`: always use GET requests</li><li>`post`: always use POST requests</li><li>`adaptive`: use a POST request if the URI length is longer than 4096 characters, otherwise use a GET request</li></ul> | `adaptive` |
+| `kroki-plantuml-include` | A file that will be included at the top of all PlantUML diagrams as if `!include file` was used. This allows you to define a common skin for your diagrams. This can be a path or a URL. |  |
```
Mogztter, line 318: Not a native speaker but I feel like "This allows you to..." is "restrictive. I think we should say that it can be used to do X but not limited to:

```suggestion
| `kroki-plantuml-include` | A file that will be included at the top of all PlantUML diagrams as if `!include file` was used. It can be useful when you want to define a common skin for all your diagrams. The value can be a path or a URL. |  |
```

ruby/lib/asciidoctor/extensions/asciidoctor_kroki/extension.rb
```diff
@@ -74,6 +75,14 @@ def process(processor, parent, attrs, diagram_type, diagram_text)
 
       private
 
+      def prepend_plantuml_config(diagram_text, diagram_type, doc)
+        if diagram_type == :plantuml && doc.attributes['kroki-plantuml-include']
+          config = File.read(doc.attributes['kroki-plantuml-include'])
```
Mogztter, line 80: I think we should use `attr` and `attr?` methods:

```suggestion
        if diagram_type == :plantuml && doc.attr?('kroki-plantuml-include')
          config = File.read(doc.attr('kroki-plantuml-include'))
```

I'm not a big fan of assigning a value in a condition but it could be reduced to:


```rb
if diagram_type == :plantuml && (plantuml_include = doc.attr('kroki-plantuml-include'))
  File.read(plantuml_include) + '\n' + diagram_text
else
  diagram_text
end
```

Mogztter: I'm a bit concerned that we have a different implementation in Ruby but at the same time we won't implement the PlantUML processor in Ruby in this pull request.

However, could you please add a "TODO" to make it clear that we need to implement the PlantUML preprocessor in Ruby and replace `File.read(plantuml_include)` by `"!include #{plantuml_include}"`?

gtudan: I don't think we should rely on "!include" before we have a preprocessor for ruby.
Wouldn't that limit the ruby implementation to files on the server and URLs, as the content of local files won't be sent to the server?

Mogztter: No worries, you are 100% correct we should not rely on `!include` until the preprocessor is implemented in Ruby but the "TODO" will help to remember this decision and what needs to be done 👌🏻

PR #124, opened 2020-07-28
remove plantuml tags as last step in processing

gtudan:
Removes all PlantUML tags (`@startuml`/`@enduml`) to mimic the behavior of the plantuml server for multiple diagrams in a file 

This is only feasible for the standard tags, as others (like `@startmindmap`) are mandatory - the diagram won't work if we remove them

### DIFF
--- a/src/preprocess.js
+++ b/src/preprocess.js
@@ -56,6 +56,24 @@ ${diagramText}
 }
 
 /**
+ * Removes all plantuml tags (@startuml/@enduml) from the diagram
+ * It's possible to have more than one diagram in a single file in the cli version of plantuml
+ * This does not work for the server, so recent plantuml versions remove the tags before processing the diagram
+ * We don't want to rely on the server to handle this, so we remove the tags in here before sending the diagram to the server
+ *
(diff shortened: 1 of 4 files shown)


Mogztter:
> Removes all PlantUML tags (@startuml/@enduml) to mimic the behavior of the plantuml server for multiple diagrams in a file
This is only feasible for the standard tags, as others (like @startmindmap) are mandatory - the diagram won't work if we remove them.

Sounds good 👍 

Regarding the build failure, we now have this intermittent failure on Windows:

```
Unable to start the browser tests suite: Invalid arguments: should be exactly one string.
```

Not sure why? 🤔

gtudan:
Yes, the windows build is flaky and seems to break for no reason - all green now.

Mogztter:
> Yes, the windows build is flaky and seems to break for no reason - all green now on 3rd try.

Yeah... probably a concurrency error but the error message is mysterious. I will add more logging to troubleshoot. Anyway it's not related to your work 😄 

The code looks good, thanks for your contribution!

PR #141, opened 2020-09-21
Lack of local includes should only warn

mwabik:
Lack of local includes should only warn and do not throw an error, because the data file can perhaps be found by kroki server.

This changes local files pre-processing to be best-effort. And reverts behavior to as in 0.8.2 version.

### DIFF
--- a/src/preprocess.js
+++ b/src/preprocess.js
@@ -195,8 +195,9 @@ function readPlantUmlInclude (url, dirPath, includeStack, vfs) {
         try {
           text = read(filePath)
         } catch (e) {
-          const message = `Preprocessing of PlantUML include failed, because reading the referenced local file '${filePath}' caused an error:\n${e}`
-          throw addCauseToError(new Error(message), e)
+          // Only warn and do not throw an error, because the data file can perhaps be found by kroki server
+          console.warn(`Skipping preprocessing of PlantUML include, because reading the referenced local file '${filePath}' caused an error:\n${e}`)
+          skip = true
         }
       }
     }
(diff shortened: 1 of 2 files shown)


Mogztter:
Thanks @mwabik 
I wonder if we should introduce an option to control this behavior?


`kroki-plantuml-unresolved-local-include`
* `warn`: log a warning
* `error`: throw an exception
* `ignore`: silently ignore

We could do the same for remote include. The default value of local include would be error and the default value for remote include would be `warn`.

In you case, you will be able to suppress the (unnecessary and thus annoying) warning.

What do you think?

 @anb0s Do you have an option on this?

mwabik:
I think, in general, the better default is to call korki server if client library has any doubts (regarding syntax, options, includes and such). Otherwise huge benefit of standardizing on kroki server can be killed by not consistent behavior between client libraries and versions. 

We are heavy intellij/plantuml (with kroki support) plugin users - and pre-processors are non-existent there anyway.  We have just issues with VS Code users, and we can't upgrade antora...  

I wonder if such config is even necessary. The  only difference is, if the error is thrown on client or server side - diagrams won't be generated either way :) There should be no impact for correct diagrams, but there will be noise in logs. So that could be only option - log level. It would also help with standard library includes.

anb0s:
We are using files with lot of local includes (own styles, macros etc.) and need traceability if files cannot be included (config path errors or typos) before sending to kroki-server to have consistent diagrams. For shared files (libraries) we are using `<std-includes>`. I think we should warn users if hard errors breaks the rendering for other use cases like referenced files available only at server (but are **not** `<std-library>`).

@Mogztter IMHO we can accept this change for now. I will ask more colleagues if we still need the hard error configurable, e.g. similar to C++ compiler options `-Wall -Werror` ==> fail-fast :)

Mogztter:
Thanks for your feedback @anb0s 
I will go ahead and merge this pull request. I will create an issue for the improvement I described.

Mogztter:
Thanks @mwabik for your contribution :tada: 
Sorry for taking so long to merge, somehow it slipped under my radar :sweat_smile:

Issue #99, opened 2020-06-16
Create more similar HTML compared to asciidoctor-diagram (width and height attributes of img element)


stenzengel:
Split of #13 to simplify tracking. The sample Asciidoctor input and the generated HTML output of asciidoctor-diagram and asciidoctor-kroki can be found there. 

This issue only considers differences in the "width" and "height" attribute:
asciidoctor-diagram sets the image "width" and "height" on the element. Kroki does not set these attributes.

Mogztter:
With https://github.com/Mogztter/asciidoctor-kroki/pull/95 if the block has explicit width and/or height attributes they will be preserved.

The remaining issue is to infer the width and height from the generated image when width and height are not explicitly defined.
In this case, Asciidoctor Diagram does the following:

**PNG**
https://github.com/asciidoctor/asciidoctor-diagram/blob/66876e9c11c401eec943ff6968cd0c5f6c443906/lib/asciidoctor-diagram/util/png.rb#L9-L22

**GIF**
https://github.com/asciidoctor/asciidoctor-diagram/blob/66876e9c11c401eec943ff6968cd0c5f6c443906/lib/asciidoctor-diagram/util/gif.rb#L10-L18

**SVG**
https://github.com/asciidoctor/asciidoctor-diagram/blob/66876e9c11c401eec943ff6968cd0c5f6c443906/lib/asciidoctor-diagram/util/svg.rb#L7-L24

Mogztter:
I've created a dedicated issue, closing.

PR #181, opened 2020-12-17
Local data file referenced relatively in Vega-Lite file

### LINKED ISSUE #98: Create more similar HTML compared to asciidoctor-diagram (alt attribute of img element)
Split of #13 to simplify tracking. The sample Asciidoctor input and the generated HTML output of asciidoctor-diagram and asciidoctor-kroki can be found there. 

This issue only considers differences in the "alt" attribute:
asciidoctor-diagram replaces the "-" character in "my-filename" with " " in the "alt" attribute. Kroki does no replace this character.

### LINKED ISSUE #114: Local data file that is referenced relatively in Vega-Lite file is looked up relatively to the Asciidoctor file
A local data file (see #53) that is referenced relatively in a Vega-Lite file is looked up relatively to the Asciidoctor file in "asciidoctor-kroki". This works differently and more logically in "asciidoctor-diagram", where the Vega-Lite file is used as origin.

### Example:
test.adoc:
```asciidoc
= Test
:kroki-fetch-diagram: true

vegalite::subdir/test.vl[]
```

subdir/test.vl:
```json
{
	"$schema": "https://vega.github.io/schema/vega-lite/v4.json",
	"data": {
		"url": "test.csv"
	},
	"encoding": {
		"x": {
			"field": "a",
			"type": "ordinal"
		},
		"y": {
			"field": "b",
			"type": "quantitative"
		}
	},
	"mark": "bar"
}
```

subdir/test.csv:
```csv
a,b
A,2.1
B,1.2
C,0.3
```

### Current behavior:
Skipping vegalite block macro. Preprocessing of Vega-Lite view specification failed, because reading the referenced local file 'test.csv' caused an error:      
Error: ENOENT: no such file or directory, open 'test.csv'

### Expected behavior:
The CSV data file should be looked up relativ to the JSON file and no error should occur.

### Workarounds:

- Put `test.csv` in the same directory as the asciidoc file
- Add relative path from asciidoc file : e.g. `"url": "subdir/test.csv"`
- Reference `test.csv` absolutely: e.g. `"url": "C:\\...\\subdir\\test.csv"`

All workarounds work with "asciidoctor-kroki" but seem to break "asciidoctor-diagram".

@Mogztter What would be the best way to fix this problem? If you have an idea, I could provide a pull request.

stenzengel:
fixes #114
fixes file URLs in Windows

Could you please have a look at this PR? 

This fixes an inconsistency if the Vega-Lite file is in a different directory than the Asciidoctor file and references a local data file. This does not fix a similar problem for PlantUML (which should probably also be fixed, as you proposed).

This fixes also a problem with file URLs, which I had in Windows (see `node-fs.js`). This could change existing behavior globally and not only for Vega-Lite diagrams!

In addition, I did some refactorings especially in the test code and moved `// @ts-node` back to the beginning of the file.

BTW: As I have some time at the moment (I regret that I did not have it earlier), I hope to fix #98 and #99 soon, which will be easier for me once this PR is merged.

### DIFF
--- a/src/asciidoctor-kroki.js
+++ b/src/asciidoctor-kroki.js
@@ -58,7 +58,7 @@ function getOption (attrs, document) {
   }
 }
 
-const processKroki = (processor, parent, attrs, diagramType, diagramText, context, baseDir) => {
+const processKroki = (processor, parent, attrs, diagramType, diagramText, context) => {
   const doc = parent.getDocument()
   // If "subs" attribute is specified, substitute accordingly.
   // Be careful not to specify "specialcharacters" or your diagram code won't be valid anymore!
@@ -67,7 +67,7 @@ const processKroki = (processor, parent, attrs, diagramType, diagramText, contex
     diagramText = parent.applySubstitutions(diagramText, parent.$resolve_subs(subs))
   }
   if (diagramType === 'vegalite') {
-    diagramText = require('./preprocess.js').preprocessVegaLite(diagramText, context, baseDir)
+    diagramText = require('./preprocess.js').preprocessVegaLite(diagramText, context, doc.getBaseDir())
   } else if (diagramType === 'plantuml') {
     const plantUmlInclude = doc.getAttribute('kroki-plantuml-include')
     if (plantUmlInclude) {
@@ -176,8 +176,7 @@ function diagramBlockMacro (name, context) {
       const diagramType = name
       try {
         const diagramText = vfs.read(target)
-        const baseDir = parent.getDocument().getBaseDir()
-        return processKroki(this, parent, attrs, diagramType, diagramText, context, baseDir)
+        return processKroki(this, parent, attrs, diagramType, diagramText, context)
       } catch (e) {
         console.warn(`Skipping ${diagramType} block macro. ${e.message}`)
         attrs.role = role ? `${role} kroki-error` : 'kroki-error'
--- a/src/asciidoctor-kroki.js
+++ b/src/asciidoctor-kroki.js
@@ -176,7 +176,7 @@ function diagramBlockMacro (name, context) {
       const diagramType = name
       try {
         const diagramText = vfs.read(target)
-        const baseDir = target.substr(0, target.lastIndexOf('/') + 1)
+        const baseDir = parent.getDocument().getBaseDir()
         return processKroki(this, parent, attrs, diagramType, diagramText, context, baseDir)
       } catch (e) {
         console.warn(`Skipping ${diagramType} block macro. ${e.message}`)
--- a/src/asciidoctor-kroki.js
+++ b/src/asciidoctor-kroki.js
@@ -58,7 +58,7 @@ function getOption (attrs, document) {
   }
 }
 
-const processKroki = (processor, parent, attrs, diagramType, diagramText, context) => {
+const processKroki = (processor, parent, attrs, diagramType, diagramText, context, baseDir) => {
   const doc = parent.getDocument()
   // If "subs" attribute is specified, substitute accordingly.
   // Be careful not to specify "specialcharacters" or your diagram code won't be valid anymore!
@@ -67,7 +67,7 @@ const processKroki = (processor, parent, attrs, diagramType, diagramText, contex
     diagramText = parent.applySubstitutions(diagramText, parent.$resolve_subs(subs))
   }
   if (diagramType === 'vegalite') {
-    diagramText = require('./preprocess.js').preprocessVegaLite(diagramText, context)
+    diagramText = require('./preprocess.js').preprocessVegaLite(diagramText, context, baseDir)
   } else if (diagramType === 'plantuml') {
     const plantUmlInclude = doc.getAttribute('kroki-plantuml-include')
     if (plantUmlInclude) {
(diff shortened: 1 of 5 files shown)


src/node-fs.js
```diff
@@ -18,7 +19,7 @@ module.exports = {
       return http.get(path, encoding)
     }
     if (path.startsWith('file://')) {
-      return fs.readFileSync(path.substr('file://'.length), encoding)
+      return fs.readFileSync(url.fileURLToPath(path), encoding)
```
Mogztter, line 22: :+1:

src/asciidoctor-kroki.js
```diff
@@ -176,7 +176,8 @@ function diagramBlockMacro (name, context) {
       const diagramType = name
       try {
         const diagramText = vfs.read(target)
-        return processKroki(this, parent, attrs, diagramType, diagramText, context)
+        const baseDir = target.substr(0, target.lastIndexOf('/') + 1)
```
Mogztter, line 179: I think we should use the built-in base directory from Asciidoctor no?

```js
const baseDir = doc.getBaseDir()
```

test/preprocess.spec.js
```diff
@@ -1,90 +1,152 @@
...
+
+  /**
+   * @param {string} diagramText
+   * @param {string} expectedPreprocessedDiagramText
+   * @param {string} [baseDir]
+   * @returns {void}
+   */
+  function expectToBeEqual (diagramText, expectedPreprocessedDiagramText, baseDir) {
+    expect(preprocessVegaLite(diagramText, {}, baseDir)).to.be.equal(expectedPreprocessedDiagramText.replace(/\r\n/g, '\n'))
+  }
```
Mogztter, line 48: Would it be possible to revert these changes? It would be easier to review :angel: 
We can still introduce them later.

Mogztter:
@stenzengel Overall it looks really good. I left a few comments but otherwise :ok_hand:

src/node-fs.js
stenzengel: Yes, the following test cases fail when running in Windows (not Linux):

- should throw an error for unexisting file referenced with "file" protocol
- should return diagramText with inlined local file referenced with "file" protocol and absolute path

Btw: I get new test failures since my last commit and even on master. Do you know why? Perhaps kroki-server changes?:

- should inline (via page attr) an SVG image with built-in allow-uri-read (available in Asciidoctor.js 2+)
- should inline (via block attr) an SVG image with built-in allow-uri-read (available in Asciidoctor.js 2+)
- should inline (via block attr overriding page attr) an SVG image with built-in allow-uri-read (available in Asciidoctor.js 2+)

src/asciidoctor-kroki.js
stenzengel: Done

test/preprocess.spec.js
stenzengel: Yes, I agree that it is very difficult to compare with the previous version. But even if I inline these helper functions, the diff would probably not be very helpful because the other changes would be too big. For example, I changed the test names, added and moved tests, extracted constants....
Could you please try to review without comparing with the previous version? If the test cases are not comprehensible, I am happy to improve them. 
Sometimes it is difficult to me to stop refactoring code ...

src/node-fs.js
Mogztter: Thanks for the clarification.

> Btw: I get new test failures since my last commit and even on master. Do you know why? Perhaps kroki-server changes?:

It should be fixed in https://github.com/Mogztter/asciidoctor-kroki/commit/a6bc1853eacc2ba9846028c6fc951b1557f9c4be (pushed yesterday on master).
Did you fetch + rebase on master ?

test/preprocess.spec.js
Mogztter: > Sometimes it is difficult to me to stop refactoring code ...

I do have the same "problem" 😃

> Could you please try to review without comparing with the previous version? If the test cases are not comprehensible, I am happy to improve them.

Fair enough, I will checkout the branch locally and let you know 👌

src/node-fs.js
stenzengel: Thanks, that solved the problem.

Mogztter:
@stenzengel I rebased on master and made two small changes, let me know if this look good to you.

stenzengel:
@Mogztter Yes, this looks better.

stenzengel:
@Mogztter Unfortunately, the bug is not fixed in the released version.  I think that the problem was introduced in commit `Use "parent.getDocument().getBaseDir()" instead of string manipulation` (30730b5d50038cf811d5b5b5c035314881aa7c41), where I did not realize that your proposition would change the semantic: Instead of using the folder of the vega-lite file, parent.getDocument().getBaseDir() seems to get the folder of the asciidoctor file. Is this correct?

The file `asciidoctor-kroki.js` is not unit tested and I did not this test it manually after this change.  I'm sorry for that.

Do you want me to open a new issue or pull request?

Mogztter:
Yes please open a new issue to describe in which context (and if possible with a reproduction case) it does not work.
Path resolution is complex and I would try to make it consistent across the Asciidoctor ecosystem.

> Instead of using the folder of the vega-lite file, parent.getDocument().getBaseDir() seems to get the folder of the asciidoctor file. Is this correct?

Yes that's true, sorry it's my bad I didn't realize we were resolving resources from the Vega file (and not from the AsciiDoc file)

> The file asciidoctor-kroki.js is not unit tested and I did not this test it manually after this change. I'm sorry for that.

No worries, we should try to add a test to make sure that it's working as expected.

PR #191, opened 2021-01-15
Issue 190: refix local data file referenced relatively

### LINKED ISSUE #190: Refix: Local data file that is referenced relatively in Vega-Lite file is looked up relatively to the Asciidoctor file
The pull request #181 does not fix issue #114: "Local data file referenced relatively in Vega-Lite file", because of the additional commit `Use "parent.getDocument().getBaseDir()" instead of string manipulation` (30730b5d50038cf811d5b5b5c035314881aa7c41)

This commit should be reverted and a test should be added to make sure that file resolution works and continues working as expected.

stenzengel:
Fixes #190 (and #114)

@Mogztter What is the best way to test that `diagramDir` is the directory of the diagram and not of the asciidoctor file? At least I renamed the variable/parameter `baseDir` to `diagramDir` to make the meaning a bit clearer. Feel free to modify the PR.


src/asciidoctor-kroki.js
```diff
@@ -176,7 +176,8 @@ function diagramBlockMacro (name, context) {
       const diagramType = name
       try {
         const diagramText = vfs.read(target)
-        return processKroki(this, parent, attrs, diagramType, diagramText, context)
+        const diagramDir = target.substr(0, target.lastIndexOf('/') + 1)
```
Mogztter, line 179: we should probably use https://nodejs.org/api/path.html#path_path_dirname_path

Mogztter:
> @Mogztter What is the best way to test that diagramDir is the directory of the diagram and not of the asciidoctor file? At least I renamed the variable/parameter baseDir to diagramDir to make the meaning a bit clearer. Feel free to modify the PR.

I think we should use `convertFile`:

```js
const registry = asciidoctor.Extensions.create()
asciidoctorKroki.register(registry)
const doc = asciidoctor.convertFile('path/to/doc.adoc', { extension_registry: registry, safe: 'unsafe' })
const html = doc.convert()
// assert that Vega-Lite `data.url` was resolved relative to the diagram file (and not the AsciiDoc file)
```

Alternatively or in addition, we could introduce a pure function that does all the preprocessing? The `processKroki` function does too much and is really an internal function. But that would require a bit of refactoring so let's first try to add a failing test case using `convertFile` (that would be fixed with your changes).

src/asciidoctor-kroki.js
Mogztter: Nevermind, I forgot we cannot use it here :(

stenzengel: Just out of curiosity. Why can `path.dirname(path)` not be used here?

stenzengel:
Thanks for the test case and the remaining renamings.

src/asciidoctor-kroki.js
Mogztter: Actually, since we are using Browserify, we could. I believe that `path.dirname` will be replaced by the following implementation:

https://github.com/browserify/path-browserify/blob/872fec31a8bac7b9b43be0e54ef3037e0202c5fb/index.js#L293-L316

But I'm fine with this "simple" implementation. We can reevaluate if someone has any issue in a browser environment.

Mogztter:
We need to do the same change for PlantUML right? If so I will create a follow-up issue to not forget.

stenzengel:
I think that this is probable, but I'm not sure, because I do not use referenced files in plantuml files.

Mogztter:
> I think that this is probable, but I'm not sure, because I do not use referenced files in plantuml files.

I just did a test and I can confirm that we have the same issue. I created #201 as a follow-up issue.

PR #195, opened 2021-01-20
Ruby: change the encoded newline to a real newline for kroki-plantuml-include

ahus1:
... as it would otherwise break the content. 

It seems that the generated URL by accident produces a SVG - although the meaning has been changed. 

Image before (wrong): 

image

Decoding the contents of the URL shows the surplus `\n` at the beginning of the line: 

```
skinparam monochrome true
\nalice -> bob: hello
```

Image after (correct): 

image


Mogztter:
Oh I see, single quotes (`'\n'`) produces literally `\n` but double quotes (`"\n"`) will produce a newline, right?

ahus1:
Yes, that's right.

PR #196, opened 2021-01-20
match "lf" line endings in .editorconfig for .gitattributes and rubocop

ahus1:
Working with an OS that is notoriously known for its CRLF line endings, I found myself in the following situation: 

* git checked out all files with CRLF
* .editorconfig asked my IDE to safe all changed files with LF
* rubocop complained about LF files, as it used the OS default

Therefore I suggest to align the settings of git and rubocop with the .editorconfig.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+# To match the .editorconfig that asks for lf only
+# Otherwise you will end up with files that show up as modified after you IDE
+* text=auto eol=lf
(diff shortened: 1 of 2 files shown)


Mogztter:
> Working with an OS that is notoriously known for its CRLF line endings, I found myself in the following situation:

The one and only :facepalm: 

> Therefore I suggest to align the settings of git and rubocop with the .editorconfig.

LGTM, thanks :ok_hand:

PR #198, opened 2021-01-21
rename attribute env-idea…

ahus1:
…to kroki-force-png as this attribute name is leaking an implementation details from an early version of this plugin.

In fact, this was only true for the JavaFX preview. The new JCEF preview supports SVG preview.


ahus1:
didn't change JavaScript elements, therefore I plead "not guilty" for the failing JavaScript builds.

ahus1:
rebased, checks are now green.

Mogztter:
> ... as this attribute name is leaking an implementation details from an early version of this plugin.

Indeed, we should rename this attribute.
What about `:kroki-force-format: png`? Since we are already using `:kroki-default-format: png`, it would be consistent.
Could you please describe this attribute in https://github.com/Mogztter/asciidoctor-kroki#configuration?

ahus1:
This would change the semantics, as now several formats would need to be supported, including the feature that the extension would need to know which format is support by what backend, see the following code comment:

```
# ... unless the diagram library does not support PNG as output format!
# Currently, mermaid, nomnoml, svgbob, wavedrom only support SVG as output format.
```

New proposal: I'd rather implement the `kroki-default-format` in the ruby variant, and drop the `kroki-force-format` altogether.

This will not be a perfect fit for the IntelliJ plugin; on the other hand the IntelliJ platform is switching to JCEF that support SVGs properly. The automatic switching to PNG caused confusion in some places before.

WDYT?

Mogztter:
> New proposal: I'd rather implement the kroki-default-format in the ruby variant, and drop the kroki-force-format altogether.

Sounds good, less is more :+1: 

> This will not be a perfect fit for the IntelliJ plugin; on the other hand the IntelliJ platform is switching to JCEF that support SVGs properly. The automatic switching to PNG caused confusion in some places before.

So your plan is to use `:kroki-default-format: png` when the IntelliJ plugin is not using JCEF?

ahus1:
Hi @Mogztter - I finally completed the rework for this. I'd be happy if you could have a look.

And yes, I plan to use `:kroki-default-format: png` when the IntelliJ plugin is not using JCEF.

Mogztter:
Looks good.

The commit message should be something like: "Remove env-idea logic (use kroki-default-format instead)", right? I can update it when merging but I want to make sure that I fully understand the change.

PR #199, opened 2021-01-21
Ruby: log via Asciidoctor logger to provide the user with a file name 

ahus1:
… and line number in the log, plus redirection of log content in IDEs.

I increased the "Metrics/AbcSize" as I didn't know a good refactoring for this method to stay below that limit. I'd appreciate a hint or help here. 

Something I found: the log line will point at...
* the first line of the block (the "----") for a block processor
* the the block macro for block macro

I started to pass down this information to KrokiProcessor.process, but then Rubocop complained about too many parameters. Then I settled with "leave it as it is" :-/ ... [see this snippet](https://github.com/asciidoctor/asciidoctor-intellij-plugin/blob/master/src/main/resources/plantuml-patch.rb#L24-L26) on how to adjust line numbers for logging.


ahus1:
rebased, not sure why JS test failed in Windows - unstable test?

Mogztter:
> rebased, not sure why JS test failed in Windows - unstable test?

Probably, we might need to adjust the timeout...

Mogztter:
I think we need to merge https://github.com/Mogztter/asciidoctor-kroki/pull/171 first (or at least we will need to update #171 if we decide to merge this change first).

If you have time to review #171 that would be great. I did these changes to allow a better integration with GitLab. I guess the Intellij IDEA plugin would also benefit from these changes but I didn't check carefully.

Mogztter:
I merged #171, we need to rebase this pull request.
If you want I can take care of it, just let me know :wink:

Mogztter:
@ahus1 Sorry I lost track but I think it's ready to go, right?

PR #200, opened 2021-01-21
adjust fixtures to latest changes in PlantUML 1.2021.0

ahus1:
... as this should fix the breaking builds on #198 and #199. 

I can re-base both PRs once this has been merged.


Mogztter:
Thanks!

Yes, it's a bit annoying... it might be better to generate PNG images and use pixelmatch to compare the expected and actual image.
But I didn't have time to do it :sweat_smile: 

Anyway, all good :+1:

ahus1:
What about .... mocking the external Kroki service that returns just the contents from the URL?

PR #211, opened 2021-02-10
Check Asciidoctor security level before reading or writing to disk

Mogztter:


Mogztter:
@anb0s @stenzengel @gtudan Just a friendly heads-up since you are using Asciidoctor Kroki preprocessing and/or `kroki-plantuml-include`

gtudan:
Thanks for the heads up. Do you happen to know what antoras default is for safe mode?

Mogztter:
> Thanks for the heads up. 

You're welcome!

> Do you happen to know what antoras default is for safe mode?

I do, Antora is using `safe`: https://gitlab.com/antora/antora/-/blob/master/packages/asciidoc-loader/lib/load-asciidoc.js#L71

PR #240, opened 2021-03-29
adding support for bitfield

am21:


am21:
bitfield just needs to be installed as a wavedrom dependency, and we can continue use `[wavedrom]` in asciidoctor.

PR #278, opened 2021-07-09
resolves #277 Post request bug correction

### LINKED ISSUE #277: Ruby asciidoctor-kroki Failed to load AsciiDoc document on post request
Launching  the command : 

```
bundle exec asciidoctor -a toc -r asciidoctor-kroki Figures.adoc -D public
```

return 

```
asciidoctor: FAILED: documentation/Figures.adoc: Failed to load AsciiDoc document - undefined method `request_post' for Net::HTTP:Class
```

demagnevalCegelec:
Fix  issue #277 
asciidoctor: FAILED: documentation/Figures.adoc: Failed to load AsciiDoc document - undefined method `request_post' for Net::HTTP:Class

### DIFF
--- a/ruby/lib/asciidoctor/extensions/asciidoctor_kroki/extension.rb
+++ b/ruby/lib/asciidoctor/extensions/asciidoctor_kroki/extension.rb
@@ -377,7 +377,7 @@ module AsciidoctorExtensions
       end
 
       def post(uri, data, _)
-        res = ::Net::HTTP.post(URI(uri), data, "Content-Type" => "text/plain")
+        res = ::Net::HTTP.post(URI(uri), data, 'Content-Type' => 'text/plain')
         res.body
       end
     end


Mogztter:
Thanks @demagnevalCegelec 
Could you please fix the style issue reported by Rubocop?

Also, I think we should probably test this method using a mock HTTP server. We have a unit test but we do not actually test the `post` method (we only check that the method is called).

ruby/lib/asciidoctor/extensions/asciidoctor_kroki/extension.rb
```diff
@@ -377,7 +377,7 @@ def get(uri, _)
       end
 
       def post(uri, data, _)
-        res = ::Net::HTTP.request_post(uri, data)
+        res = ::Net::HTTP.post(URI(uri), data, "Content-Type" => "text/plain")
```
Mogztter, line 380:
```suggestion
        res = ::Net::HTTP.post(URI(uri), data, 'Content-Type' => 'text/plain')
```

PR #294, opened 2021-09-11
Update Rubocop and other gems + fix errors

archiloque:
When working on the project I have some errors because I'm using Ruby 3.0 and the Rubobop version used by the project is not aware of this version so I get an error message  (`Supported versions: 2.3, 2.4, 2.5, 2.6, 2.7`).

I have update Rubocop and took the opportunity to update the other gems at the same time since they don't have any impact.

After running the auto fix, I manually fixed the issues that have a low impact on the code and disable checks in the files for the other ones, the number of changes seems reasonable.

I also enabled the new rules for Rubocop as suggested bu the doc.

I also added the new `rubocop-rake` gem since it has no impact but I _didn't_ added `rubocop-rspec` because it detects a lot of violations which should have a big impact on the code.

If you want I can add it and fix the violations, I wanted to know what you prefer before doing the required work, let me know what you prefer.


ruby/.rubocop.yml
```diff
@@ -1,10 +1,17 @@
+require:
+  - rubocop-rake
```
Mogztter, line 2: What does `rubocop-rake` does? It does static analysis on the Rakefile?

Mogztter:
Looks good, I would rather leave `rubocop-rake` out for now. I don't think we need it, one less dependency, one less thing to worry about 😄

ruby/.rubocop.yml
archiloque: Yes, there are specific rules for rake files, I will remove it

archiloque:
> Looks good, I would rather leave `rubocop-rake` out for now. I don't think we need it, one less dependency, one less thing to worry about 😄

I've removed it, I also took the opportunity to bump the required Ruby version to make it match the one I added in `ruby/.rubocop.yml` (I forgot I added it), tell me if you prefer to keep the 2.6 version.

ruby/.rubocop.yml
Mogztter: Thanks

Mogztter:
> I've removed it, I also took the opportunity to bump the required ruby/.ruby-version to make it match the one I added in ruby/.rubocop.yml (I forgot I added it), tell me if you prefer to keep the 2.6 version.

Perfect, I'm fine with Ruby 2.7.
Ruby 2.6 is already in "security maintenance", so it's probably better to use Ruby 2.7 as the minimum version:

https://www.ruby-lang.org/en/downloads/branches/

Mogztter (approved):
Thanks 👍🏻

PR #301, opened 2021-10-16
Register structurizr as a new Kroki diagram

jagedn:
add structurizr as a new diagram supported by kroki


Mogztter:
Looking good, thanks @jagedn 
If build comes back green, I will go ahead and merge 🎉

Issue #308, opened 2021-11-14
Fix how to soft set a kroki attribute in Antora (use a trailing @ on the value, not the attribute)


sturtison:
Hi @Mogztter while reading the README.md, following and using the guide, it seems that the info below is perhaps out of date.

https://github.com/Mogztter/asciidoctor-kroki/blob/0cfab7d60eeccb6bf64b17928156054ea6c799a9/README.md?plain=1#L361-L367

The [antora docs for soft set](https://docs.antora.org/antora/2.3/component-attributes/#soft-set) describe that the @ should trail the value, not the attribute.

The actual result is line should hold ```kroki-default-format: png@```

I'm raising a PR from my fork using this issue number if it's suitable for you to accept.

/edits: Format result line.

Mogztter:
fixed in #309

PR #309, opened 2021-11-14
fix(README.md): #308 Soft set kroki attribute

sturtison:
Put the trailing @ on the value instead of the attribute.

To fix issue #308


Mogztter:
Thanks!

Issue #346, opened 2022-02-18
OS-specific "kroki-plantuml-include-paths" directory-delimiter inhibits cross-platform collaboration


thepartisan:
The current behavior of the setting "kroki-plantuml-include-paths" is to have one string to define multiple include paths separated by ";" on Windows and ":" on nix-OSes (see [documentation](https://github.com/Mogztter/asciidoctor-kroki#configuration) ). This inhibits sharing one workspace settings.json between people that use different OSes.

A non-backwards-compatibility-breaking fix could be to additionally allow "," as separator.

Mogztter:
> A non-backwards-compatibility-breaking fix could be to additionally allow "," as separator.

One issue is that `,` is a valid character on Unix: `foo,`.

> This inhibits sharing one workspace settings.json between people that use different OSes.

That would be beneficial, could you please share an example of such configuration? Are you using absolute paths? Are you using a shared drive?

Mogztter:
Closing.
If you are still interested please join https://kroki.zulipchat.com/ so we can discuss the best approach.

PR #347, opened 2022-02-18
Allow comma as path delimiter for include paths

thepartisan:
This fixes the problem of platform-dependency of the
`kroki-plantuml-include-paths` attribute.
In the old implementation it was not possible to share
a configuration with multiple include paths defined
between Windows and Unix users.

Fixes Mogztter/asciidoctor-kroki#346


Mogztter:
Closing, feel free to suggest an alternative solution in https://github.com/Mogztter/asciidoctor-kroki/issues/346

PR #372, opened 2022-06-04
resolves #371 consult :to_dir option on document to resolve output directory

### LINKED ISSUE #371: Output directory not honoring value of :to_dir option
To be consistent with the behavior of Asciidoctor Diagram, when the `imagesoutdir` attribute is not specified, Asciidoctor Kroki should save images to the `imagesdir` folder relative to the output directory of the processor (the directory where the output file is being written). In code, the location of the output directory can be expressed as:

```rb
(doc.attr 'outdir') || (doc.nested? ? doc.parent_document : doc).options[:to_dir]
```

This uses the value of `outdir`, if set, falling back to the `:to_dir` option if it's not set.

Currently, Asciidoctor Kroki is instead looking for the `to_dir` attribute:

```rb
elsif (to_dir = doc.attr('to_dir'))
```

`to_dir` is not a recognized built-in attribute name in Asciidoctor. Therefore, I think this code is incorrect.

mojavelinux:


Mogztter:
Thanks!
We should probably also update the JavaScript code:

https://github.com/Mogztter/asciidoctor-kroki/blob/4c9cb535ec6747f2e76d088944451114957002e3/src/fetch.js#L7

I will try to add a test case on both Ruby and JavaScript to make sure that the `to_dir` option is honored.

ruby/lib/asciidoctor/extensions/asciidoctor_kroki/extension.rb
```diff
@@ -252,10 +252,8 @@ def output_dir_path(doc)
         images_dir = doc.attr('imagesdir', '')
         if (images_output_dir = doc.attr('imagesoutdir'))
           images_output_dir
-        elsif (out_dir = doc.attr('outdir'))
+        elsif (out_dir = doc.attr('outdir') || (doc.nested? ? doc.parent_document : doc).options[:to_dir])
```
mojavelinux, line 255: I've patched core so extensions don't have to worry about the nested document, if you prefer to wait for a release. See https://github.com/asciidoctor/asciidoctor/commit/7edc9da023522be67b17e2a085d72e056703a438

Mogztter: I think it's safer to keep it a bit longer but we should certainly add a comment to state that this is unnecessary with Asciidoctor >= 2.0.18

PR #380, opened 2022-07-29
resolves #379 support windows EOL line endings

### LINKED ISSUE #379: Include in Activity Diagram does not work ( due to "\r" )
Hello,
first of all thanks for this amazing tool. It is very helpful. 

Today I came across an issue on Windows where I try to include a stylesheet in an activity diagram.
It it related to #201. 
As I have found out, the path to the included file gets not resolved due to "\r" at line end of include reference.

Error message :  

' caused an error:sing of PlantUML include, because reading the referenced local file 'stylesheet.iuml
'rror: ENOENT: no such file or directory, open 'stylesheet.iuml

Please see attached sample project.

asciidoctor-kroki_cr-issue_01

The issue can be tracked down in `node_modules/asciidoctor-kroki/src/preprocess.js` line 125 as shown in above screen shot.

Maybe it is not the best place to fix it as shown in above screen shot but at least with the modification there it works.
Could you please check and maybe fix it where appropriate?

Thanks and Best Regards

Sample Project:
[adoc-web_cr-issue.zip](https://github.com/Mogztter/asciidoctor-kroki/files/9132190/adoc-web_cr-issue.zip)

mijlouis:
When a diagram is using Windows EOL line endings (\r\n) and using include
of a stylesheet, an error message is thrown reporting that the file to
include cannot be found (because of a remaining \r on the url of the
code.)
This fix adresses this issue and removes any trailing spaces, \r,\n,\r\n
in the resolved include file.

### DIFF
--- a/src/preprocess.js
+++ b/src/preprocess.js
@@ -124,7 +124,7 @@ function preprocessPlantUmlIncludes (diagramText, dirPath, includeOnce, includeS
           const target = parseTarget(args[1])
           const urlSub = target.url.split('!')
           const trailingContent = target.comment
-          const url = urlSub[0].replace(/\\ /g, ' ')
+          const url = urlSub[0].replace(/\\ /g, ' ').replace(/(\r\n|\n|\r|\s*)$/mg, '')
           const sub = urlSub[1]
           const result = readPlantUmlInclude(url, [dirPath, ...includePaths], includeStack, vfs, logger)
           if (result.skip) {
--- a/src/preprocess.js
+++ b/src/preprocess.js
@@ -124,7 +124,7 @@ function preprocessPlantUmlIncludes (diagramText, dirPath, includeOnce, includeS
           const target = parseTarget(args[1])
           const urlSub = target.url.split('!')
           const trailingContent = target.comment
-          const url = urlSub[0].replace(/\\ /g, ' ').replace(/(\r\n|\n|\r|\s*)$/mg, '')
+          const url = urlSub[0].replace(/\\ /g, ' ').replace(/\s+$/g, '')
           const sub = urlSub[1]
           const result = readPlantUmlInclude(url, [dirPath, ...includePaths], includeStack, vfs, logger)
           if (result.skip) {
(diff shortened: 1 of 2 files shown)


src/preprocess.js
```diff
@@ -124,7 +124,7 @@ function preprocessPlantUmlIncludes (diagramText, dirPath, includeOnce, includeS
           const target = parseTarget(args[1])
           const urlSub = target.url.split('!')
           const trailingContent = target.comment
-          const url = urlSub[0].replace(/\\ /g, ' ')
+          const url = urlSub[0].replace(/\\ /g, ' ').replace(/(\r\n|\n|\r|\s*)$/mg, '')
```
Mogztter, line 127: I believe it can be simplified to:

```suggestion
          const url = urlSub[0].replace(/\s/g, ' ')
```

mijlouis: Hello @Mogztter ,
when you change to what you have suggested, there are several tests that are failing.

image

image

Best Regards

Mogztter: Indeed, we should only remove trailing space characters:

```js
urlSub[0].replace(/\s+$/g, ' ')
```

Mogztter: I don't remember exactly why we need `.replace(/\\ /g, ' ')` but I think we should keep it for now

mijlouis: Hello @Mogztter,
I changed as per your advice to 
`const url = urlSub[0].replace(/\\ /g, ' ').replace(/\s+$/g, '')`
and did some more tests ... also on the project where I identified the issue and where I have the complete context.
All tests passed successfully.
Thanks and Best Regards

mijlouis: Hello @Mogztter,
sorry for asking since I never contributed to an open source project via pull request.
What is the next thing I need to do?
Is there any further action needed by my side?

Thanks

Mogztter: Hello @mijlouis no it's all good now, I will merge it, thanks for your contribution 🤗

Issue #288, opened 2021-08-19
Can't open included iuml when running Antora


jvmlet:
I'm generating the docs with Antora , `myDiagramm.puml` looks like below :

````yml
@startuml
!include globals.iuml
....

@enduml
````

and `globals.iuml` resides in the same folder as `myDiagramm.puml`, which is `partials/uml`

When executing Antora, I'm getting
```
Skipping preprocessing of PlantUML include, because reading the referenced local file 'globals.iuml' caused an error:
Error: ENOENT: no such file or directory, open 'globals.iuml'
```

**Comment by Mogztter** [comment]
> and globals.iuml resides in the same folder as myDiagramm.puml, which is partials/uml

Technically, this is not the correct location.
From the Antora documentation:

> Partials are resources that contain content snippets, such as common descriptions, terminology, or reference tables, that you plan to reuse in one or more pages. **Partials are saved as AsciiDoc source files (.adoc) and stored in a partials directory.**

https://docs.antora.org/antora/2.3/partials-directory/#partials

I think you should store them in `examples`: https://docs.antora.org/antora/2.3/examples-directory/

Anyway, could you please share how do you reference `myDiagramm.puml` in your AsciiDoc file?

**Comment by Mogztter** [comment]
It won't work if you are referencing your diagram as follows:


**modules/ROOT/examples/hello.puml**
```puml
@startuml
!include globals.iuml

Guillaume -> Alexander : Hello!
@enduml

```

**modules/ROOT/examples/globals.iuml**
```puml
skinparam monochrome true
skinparam ranksep 20
skinparam dpi 150
skinparam arrowThickness 0.7
skinparam packageTitleAlignment left
skinparam usecaseBorderThickness 0.4
skinparam defaultFontSize 12
skinparam rectangleBorderThickness 1
```

```adoc
[plantuml]
----
include::example$hello.puml[]
----

jvmlet:
Thanks for your replay, @Mogztter 
I can't use alternative approach , because `hello.puml` won't be rendered by IDE
Setting `kroki-plantuml-include-paths`  also  fails with the same error :
````
asciidoc:
  attributes:
    kroki-plantuml-include-paths: './docs/modules/ROOT/examples'
````
I'm using `plantuml::{examplesdir}/myDiagramm.puml[]` from the `adoc` file
Both `myDiagramm.puml` and `globals.iuml` are under  `ROOT/examples` dir

Mogztter:
[...]

PR #396, opened 2022-12-13
fix dirname implementation when Antora support is enabled

ccantill:
There's an issue with Antora, where using a diagram from a file that uses !include the included file can not be resolved. That's because it falls back to the default filesystem "exists" call where it should resolve the file using antora's catalog. This PR fixes that.

Fixes https://github.com/Mogztter/asciidoctor-kroki/issues/288


Mogztter:
@ccantill could you please share an example? Are you using an Antora resource ID in the PlantUML `!include` directive?

ccantill:
There's only a resource ID in the include in asciidoc. But there are some issues resolving the include path from there. This PR fixes one of those issues.

docs/modules/ROOT/pages/index.adoc:
```
plantuml::partial$./ab.puml[opts=interactive]
```

docs/modules/ROOT/partials/ab.puml:
```
alice -> bob
!include ba.puml
```

docs/modules/ROOT/partials/ba.puml:
```
bob -> alice
```

As you can see, it still requires to put a `./` after the `partial$` to make it work, because the `diagramDir` is resolved by taking everything before the last `/`. If you would use `plantuml::partial$ab.puml[opts=interactive]` then diagramDir becomes empty and resolution is broken again. I'm not sure if it's acceptable to change that to split on both `$` and `/`.

Mogztter:
Oh I see!
I think we should provide a specific implementation to compute `diagramDir` when the target is an Antora resource ID.

The following code assumes that we are working with a POSIX path:

```js
const diagramDir = target.substr(0, target.lastIndexOf('/') + 1)
```

Instead, we should probably introduce a new method in the Virtual FileSystem "Interface":

```js
module.exports = {
 dirname: (path) => {
    // ...
 }
}
```

ccantill:
I'll add that to the PR. Is there a trick to get the tests running again? I'm not having much luck running them from master.

Mogztter:
> I'll add that to the PR. Is there a trick to get the tests running again? I'm not having much luck running them from master.

We are running tests against https://kroki.io/. Since a new version has been deployed, assertions need to be updated (i.e., some diagrams are slightly different).

I might require `docker` to run the test suite in order to test against a fixed version. To fix tests on master you will need to update the expected with the actual value (you can do a quick diff between the SVG to make sure that nothing changed drastically)

ccantill:
Fixed the tests and the directory resolution, and added a test for that use case.

Mogztter:
Thank you 🙌🏻

src/node-fs.js
```diff
@@ -22,5 +22,8 @@ module.exports = {
       return fs.readFileSync(url.fileURLToPath(path), encoding)
     }
     return fs.readFileSync(path, encoding)
+  },
+  dirname: (resourceId) => {
+    return resourceId.substring(0, Math.max(resourceId.lastIndexOf('/'), resourceId.lastIndexOf('\\')) + 1)
```
Mogztter, line 27: we can use `path.dirname`: https://nodejs.org/api/path.html#pathdirnamepath

ccantill: I was thinking that, but it depends on what happens to that resourceId before it reaches that function. In any case, it still didn't fix the issue on the Windows build. I'll need to set up a Windows environment to test and debug it locally.

src/antora-adapter.js
```diff
@@ -29,6 +29,13 @@ module.exports = (file, contentCatalog, vfs) => {
     read: (resourceId, format) => {
       const target = contentCatalog.resolveResource(resourceId, file.src)
       return target ? target.contents.toString() : baseReadFn(resourceId, format)
+    },
+    exists: (resourceId) => {
+      const target = contentCatalog.resolveResource(resourceId, file.src)
+      return target !== undefined
+    },
+    dirname: (resourceId) => {
+      return resourceId.substring(0, Math.max(resourceId.lastIndexOf('$'), resourceId.lastIndexOf('/'), resourceId.lastIndexOf('\\')) + 1)
```
Mogztter, line 38: This is unreliable and we should instead use the `contentCatalog` to resolve the resource. I believe it's possible to build an absolute `dirname` path from the resource returned by Antora.

ccantill: Could be... but I think that'll require a lot more refactoring than I'm comfortable with doing in this code base I'm afraid.

src/node-fs.js
Mogztter: In theory, `path.dirname` should use `path.sep` which is OS depend so it should work fine on Linux or Windows.

src/antora-adapter.js
ccantill: Ok, I see now. The issue is then that the file resolution is a bit all over the place. The `antora-adapter` does path resolution already, but the `node-fs` one doesn't; that's handled in other places but that breaks the resolution that's in `antora-adapter`. I guess the solution is to add resolution to `node-fs` and removing it from the other places.

ccantill: > This is unreliable and we should instead use the `contentCatalog` to resolve the resource. I believe it's possible to build an absolute `dirname` path from the resource returned by Antora.

Sorry, after two weeks of holidays I forgot how it worked, but this isn't there to create an absolute path. Instead it resolves the path from the PlantUML `!include` statement to a resourceId. It's still antora that's reading it from the catalog.

Mogztter: We are using the Virtual File System abstraction to get information about a resource. When we are using this extension through Antora we need to use the content catalog.

Antora knows where a resource is located and can compute a dirname. We need an absolute diagram directory path since PlantUML `!include` relative path will be resolved using this diagram directory.


**/path/to/modules/ROOT/pages/doc.adoc**
```
plantuml::example$architecture/diagram.puml[]
```

**/path/to/modules/ROOT/examples/architecture/diagram.puml**
```
!include ../styles.puml

Alice -> Bob: Hello
```

**/path/to/modules/ROOT/examples/styles.puml**
```
skinparam monochrome true
skinparam ranksep 20
skinparam dpi 150
skinparam arrowThickness 0.7
```

Antora knows that `example$architecture/diagram.puml` is located in `/path/to/modules/ROOT/examples/architecture`. If we want to resolve `../styles.puml` we need to compute the dirname (using Antora):

```
/path/to/modules/ROOT/examples/architecture/../styles.puml -> /path/to/modules/ROOT/examples/styles.puml
```

Does it make sense? Am I missing something?

ccantill: How I read the antora-adapter is that it doesn't resolve the resource ID to a file path to be read, but it returns the actual contents of the resource: https://github.com/Mogztter/asciidoctor-kroki/blob/master/src/antora-adapter.js#L31

We can't resolve the included file relative to the diagrams file path because we don't have it; we only have the Antora resource ID which is the "example$architecture/diagram.puml", so we have to find a resource ID for the included file by getting the "dirname" from the diagram resource ID and resolving the included file through that so the content catalog can provide it.

The reason I think why it wasn't working on Windows is that the path joining used `path.join` which on Windows will use backslash instead of forward slash, and I suspect Antora doesn't like that. I've committed a change that should fix it for real this time. I don't have a running Windows dev environment to test though.

src/preprocess.js
```diff
@@ -176,9 +176,10 @@ function preprocessPlantUmlIncludes (diagramText, dirPath, includeOnce, includeS
  */
 function resolveIncludeFile (includeFile, includePaths, vfs) {
   const exists = typeof vfs !== 'undefined' && typeof vfs.exists === 'function' ? vfs.exists : require('./node-fs.js').exists
+  const join = typeof vfs !== 'undefined' && typeof vfs.join === 'function' ? vfs.join : path.join
   let filePath = includeFile
   for (let i = 0; i < includePaths.length; i++) {
-    const localFilePath = path.join(includePaths[i], includeFile)
+    const localFilePath = join(includePaths[i], includeFile)
     if (exists(localFilePath)) {
```
Mogztter, line 183: I think this part is wrong, we should resolve file using the file system (even when we are using Antora). The reason is that the PlantUML `!include` directive does not handle Antora resource IDs.

That's probably why you had to introduce `vfs.join`.

Path resolution is complicated especially when the context impacts how we should resolve a path.

ccantill: This is there for the case where we join the antora resource dirname (e.g. "partial$architecture") with the included relative path (e.g. "ba.puml"), where Antora requires forward slashes in resource ids (i.e. "partial$architecture/ba.puml" and not "partial$architecture\ba.puml").

ccantill: What you're saying is completely rewriting the way the Antora resources are loaded. Honestly, I'm not sure that's worth the effort, and it's not an effort I can afford to put into this right now. I can close the pull request and just continue running off of our own fork if you like.

Mogztter: Are you using Antora resource IDs in PlantUML files with the `!include` directive?

ccantill: No, I'm using relative paths in !include directives. See the test case at https://github.com/Mogztter/asciidoctor-kroki/pull/396/files#diff-f387f1c7527cf78a5c58201fcd970bbaaa8dbd7a9dbfb9b51e58b69f589226ac

ccantill: Looking further at the Antora documentation, I don't see any other way to do it. https://docs.antora.org/antora/latest/organize-content-files/#antoras-virtual-filesystem states:
> The only aspect of a file that maps back to the location on the filesystem is the family-relative path

Mogztter: I will take a deeper look later this week, maybe it's more complicated than what I have in mind.

ccantill: @ggrossetie Any update on this?

PR #415, opened 2023-05-22
Add Ruby 3.2 to CI

digitalmoksha:


digitalmoksha:
@ggrossetie I think this is ready for review

.github/workflows/build-ruby.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest]
-        ruby: ['2.7', '3.0', '3.1']
+        ruby: ['2.7', '3.0', '3.1', '3.2']
```
ggrossetie, line 16:
```suggestion
        ruby: ['2.7', '3.2']
```

I don't think we need to test on every minor versions of Ruby so 2.7 and 3.2 should be enough

PR #414, opened 2023-05-22
Add symbolator

zebreus:
This PR adds support for creating symbolator diagrams

PR #417, opened 2023-06-21
Recreate lockfile

zebreus:
I had some issues with multiple different versions of the path-exists dependency when trying to package asciidoctor-kroki for nix. Deleting and recreating the lockfile solved that issue, it also reduced the number of dependencies with high severity vulnerabilities from 12 to 9.

This PR was originally #414 . That one was accidentally created from my main branch instead of the feature branch. GitHub does not allow changing the source branch of a PR so this is a new one with the correct changes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "asciidoctor-kroki",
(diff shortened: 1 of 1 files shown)


ggrossetie:
I've updated the `package-lock.json`, it should be OK now.

PR #429, opened 2023-08-13
Handle non-200 error codes separately

sransara:
Closes: https://github.com/ggrossetie/asciidoctor-kroki/issues/430

Handle non-200 error codes separately to give more visibility when the server is having issues.

With this the error messages show something like:
```
Skipping mermaid block. GET https://... - server returns 500 status code
```


ggrossetie:
Thanks for your contribution.
Before we proceed, could you please open an issue to discuss why this change is needed? (i.e., in which case you are experiencing an odd behavior, what is the current behavior, what is the expected behavior...).

sransara:
@ggrossetie sure thing. I have created and linked the issue.

ggrossetie:
superseded by #432

Issue #440, opened 2023-12-04
wireviz block type is rendered as a literalblock with ruby gem v0.9.1


lomax:
Trying to render the wireviz block from example 
```
[wireviz]
....
connectors:
  X1:
    type: D-Sub
    subtype: female
    pinlabels: [DCD, RX, TX, DTR, GND, DSR, RTS, CTS, RI]
  X2:
    type: Molex KK 254
    subtype: female
    pinlabels: [GND, RX, TX]

cables:
  W1:
    gauge: 0.25 mm2
    length: 0.2
    color_code: DIN
    wirecount: 3
    shield: true

connections:
  -
    - X1: [5,2,3]
    - W1: [1,2,3]
    - X2: [1,3,2]
  -
    - X1: 5
    - W1: s
....
```
with ruby gem v0.9.1
`asciidoctor -r asciidoctor-kroki wireviz.adoc`
results in _literalblock_ instead of _imageblock kroki_.

Is there a reason why [`SUPPORTED_DIAGRAM_NAMES`](https://github.com/asciidoctor/asciidoctor-kroki/blob/872ea774b4925e3e050d234bb802e55f8bf9cabb/ruby/lib/asciidoctor/extensions/asciidoctor_kroki/extension.rb#L108C5-L108C28) is not altered in #423?

As [v0.9.1 release](https://github.com/asciidoctor/asciidoctor-kroki/releases/tag/ruby-v0.9.1) mentioned **Wireviz** support shouldn't CI catch this issue?

PR #423, opened 2023-07-14
add wireviz support

brendanhaines:

PR #441, opened 2023-12-04
Add wireviz support to the ruby gem

lomax:
Add **wireviz** to the list of supported block types for the Ruby Gem. This were missing in #423.

Fixes functional part of #440. Tests still needed.


ggrossetie:
Thanks !

PR #452, opened 2024-06-12
‌Bugfix: fix kroki-fetch-diagram attribute in JS version for NodeJS

AshkanV:
Us node-xmlhttprequest instead of unxhr and also add support for using files in disk cache if available

### DIFF
--- a/dist/browser/asciidoctor-kroki.js
+++ b/dist/browser/asciidoctor-kroki.js
@@ -20057,7 +20057,7 @@ function extend() {
 },{}],70:[function(require,module,exports){
 module.exports={
   "name": "asciidoctor-kroki",
-  "version": "0.17.0",
+  "version": "0.18.1",
   "description": "Asciidoctor extension to convert diagrams to images using Kroki",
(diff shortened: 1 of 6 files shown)


ggrossetie:
Thanks for your pull request. Could you please explain this change: "Use node-xmlhttprequest instead of unxhr"? `unxhr` is a fork of `node-xmlhttprequest` with bug fixes and improvements since the later is unmaintained.

Also, could you please add a test for this fix?

ggrossetie:
I'm confused, why did you close this pull request?

AshkanV:
Sorry I didn't know that `unxhr` is a the better option, as there was some debates which says `unxhr` cant be run in VSCode. So I think there are two issues, one with the library and one with implementation of the attribute.

So I deleted my branch and will commit a clean one in a new branch and make another pull request.

By the way I am C++ developer and didn't find any doc in this repository on how to create a test and I myself is not familiar with the standard ways and workflows of a Node project. Could you please show me a clue on how to write a test for this project.

Thanks

PR #453, opened 2024-06-12
bugfix/fix-kroki-fetch-diagram-attribute-in-js-version

AshkanV:
Fixes the kroki-fetch-diagram attribute in NodeJS Version and also use the disk cache file if available

### DIFF
--- a/dist/browser/asciidoctor-kroki.js
+++ b/dist/browser/asciidoctor-kroki.js
@@ -20057,7 +20057,7 @@ function extend() {
 },{}],70:[function(require,module,exports){
 module.exports={
   "name": "asciidoctor-kroki",
-  "version": "0.17.0",
+  "version": "0.18.1",
   "description": "Asciidoctor extension to convert diagrams to images using Kroki",
   "type": "commonjs",
   "main": "./src/asciidoctor-kroki.js",
--- a/src/asciidoctor-kroki.js
+++ b/src/asciidoctor-kroki.js
@@ -64,7 +64,7 @@ const createImageSrc = (doc, krokiDiagram, target, vfs, krokiClient) => {
   const shouldFetch = doc.isAttribute('kroki-fetch-diagram')
   let imageUrl
   if (shouldFetch && doc.getSafe() < SAFE_MODE_SECURE) {
-    imageUrl = require('./fetch.js').save(krokiDiagram, doc, target, vfs, krokiClient)
+    imageUrl = require('./fetch.js').save(krokiDiagram, doc, target, vfs, krokiClient, true)
   } else {
     imageUrl = krokiDiagram.getDiagramUri(krokiClient.getServerUrl())
   }
--- a/src/fetch.js
+++ b/src/fetch.js
@@ -21,7 +21,7 @@ const getOutputDirectory = (doc) => {
   return baseDir
 }
 
-module.exports.save = function (krokiDiagram, doc, target, vfs, krokiClient) {
+module.exports.save = function (krokiDiagram, doc, target, vfs, krokiClient, use_data_url = false) {
   const exists = typeof vfs !== 'undefined' && typeof vfs.exists === 'function' ? vfs.exists : require('./node-fs.js').exists
   const read = typeof vfs !== 'undefined' && typeof vfs.read === 'function' ? vfs.read : require('./node-fs.js').read
   const add = typeof vfs !== 'undefined' && typeof vfs.add === 'function' ? vfs.add : require('./node-fs.js').add
@@ -31,6 +31,7 @@ module.exports.save = function (krokiDiagram, doc, target, vfs, krokiClient) {
   const format = krokiDiagram.format
   const diagramName = `${target || 'diag'}-${rusha.createHash().update(diagramUrl).digest('hex')}.${format}`
   const filePath = path.format({ dir: imagesOutputDirectory, base: diagramName })
+
   let encoding
   let mediaType
   if (format === 'txt' || format === 'atxt' || format === 'utxt') {
@@ -43,13 +44,26 @@ module.exports.save = function (krokiDiagram, doc, target, vfs, krokiClient) {
     mediaType = 'image/png'
     encoding = 'binary'
   }
+  
   // file is either (already) on the file system or we should read it from Kroki
-  const contents = exists(filePath) ? read(filePath, encoding) : krokiClient.getImage(krokiDiagram, encoding)
-  add({
-    relative: imagesOutputDirectory,
-    basename: diagramName,
-    mediaType,
-    contents: Buffer.from(contents, encoding)
-  })
-  return diagramName
+
+  let contents
+
+  if(!exists(filePath)) {
+    contents = read(diagramUrl)
+
+    add({
+      relative: imagesOutputDirectory,
+      basename: diagramName,
+      mediaType,
+      contents: Buffer.from(contents, encoding)
+    })
+  }
+
+  if (use_data_url) {
+    if (!contents) contents = read(filePath)
+    return 'data:' + mediaType + ";base64," + Buffer.from(contents, encoding).toString('base64')
+  } else {
+    return diagramName
+  }
 }
--- a/src/http/http-client.js
+++ b/src/http/http-client.js
@@ -4,8 +4,10 @@ const httpRequest = (XMLHttpRequest, uri, method, headers, encoding = 'utf8', bo
   try {
     const xhr = new XMLHttpRequest()
     xhr.open(method, uri, false)
-    for (const [name, value] in Object.entries(headers)) {
-      xhr.setRequestHeader(name, value)
+    if (headers) {
+      for (const [name, value] in Object.entries(headers)) {
+        xhr.setRequestHeader(name, value)
+      }
     }
     if (encoding === 'binary') {
       xhr.responseType = 'arraybuffer'


ggrossetie:
> Sorry I didn't know that unxhr is the better option, as there was some debates which says unxhr cant be run in VSCode. So I though there are two issues, one with the library and one with implementation of the attribute.

I believe that `node-xmlhttprequest` will have the same issue since it's basically the same codebase and use the same mechanism to execute synchronous HTTP requests.

> So I deleted my branch and will commit a clean one in a new branch and make another pull request.

👍🏻 

> By the way I am C++ developer and didn't find any doc in this repository on how to create a test and I myself is not familiar with the standard ways and workflows of a Node project. Could you please show me a clue on how to write a test for this project.

No worries.
In that case, could you describe a specific case (AsciiDoc file + environment) that currently doesn't work?

AshkanV:
Thank you
This is should be a fix for:
https://github.com/asciidoctor/asciidoctor-vscode/issues/683 
and probably:
https://github.com/asciidoctor/asciidoctor-kroki/issues/397

the sample document would be:

```adoc
= Title
:safe: unsafe
:data-uri: 
:allow-uri-read:
:kroki-fetch-diagram:

== Subtitle

[plantuml]
----
@startuml
Bob -> Alice : Request
Alice -> Bob : Response
@enduml
----
```

The only setting applied is enabling kroki.

if I replace `:kroki-fetch-diagram:` with `:kroki-fetch-diagram!:` the diagram is showing

ggrossetie:
I don't think it will fix this issue.
The reason `kroki-fetch-diagram` option does not work in VS code is because `unxhr` (and `node-xmlhttprequest`) are executing a Node script synchronously using:

```js
spawn(process.argv[0], ["-e", execString])
```


https://github.com/driverdan/node-XMLHttpRequest/blob/97966e4ca1c9f2cc5574d8775cbdacebfec75455/lib/XMLHttpRequest.js#L507

```js
require('child_process').execSync(`"${process.execPath}" ...`)
```
https://github.com/ggrossetie/unxhr/blob/4fbc39978b99180c24cd31ac802c2b4ae5dca3c6/lib/XMLHttpRequest.js#L520C60-L524

It works relatively well in a Node environment, since `process.execPath` (or `process.argv[0]`) will resolve to the path to the `node` binary.
Unfortunately, when running in VS code, Node is somehow embedded and cannot be used to run scripts. So `process.execPath` will resolve to the binary `code` which cannot execute Node scripts.

See https://github.com/ggrossetie/unxhr/issues/98

I haven't try Enrik Berkhan solution but I believe it's the only way to fix this issue.

AshkanV:
Thank you for your detailed description.

But after the changes in this merge request I was able to retrieve diagram files and have them embedded in the final HTML output and the export is single standalone HTML file again. Currently I am using these modifications on my local VSCode AsciiDoc plugin.

Before these changes, I needed to export using the Ruby version of AsciiDoctor to have an standalone HTML file.

ggrossetie:
I can confirm that `unxhr` seems to be working on the latest version of VS code. I can retrieve the image from Kroki and the file is also written on the disk:

image

The issue is that the file is written at the root of the repository but we try to load the image from the "output directory".
If I do something like (don't do that!):

image

Then it's working!

Anyway, what we should do is embed images as data URI if `data-uri` is set. 

As a next step, we should try to find a solution to make it work in preview environment (such as VS code) when `data-uri` is not set.

ggrossetie:
Since we cannot use `data-uri` in VS code, I went for `kroki-data-uri`:

image

AshkanV:
Thank you

As a matter of fact I always set the `imagesoutdir` to the temp folder of the OS and also set an attribute in the setting of the VSCode plugin for preview and check it using AsciiDoc syntax:

```ad
ifdef::vscode-preview[]
:kroki-fetch-diagram!:
endif::[]
ifndef::vscode-preview[]
:kroki-fetch-diagram:
endif::[]
```
So I get best performance in preview.

AshkanV:
I think if I don't set the `imagesoutdir` the Ruby version also saves the file in the root directory of the document.

> Since we cannot use `data-uri` in VS code, I went for `kroki-data-uri`:

Oh, I didn't know that data-uri is not working in VSCode. Because I did use it before to embed external images inside final HTML export and if I didn't use it, the result HTML just keep a URL to its location

ggrossetie:
> Oh, I didn't know that data-uri is not working in VSCode. Because I did use it before to embed external images inside final HTML export and if I didn't use it, the result HTML just keep a URL to its location

Reading local files from VS code is not supported by Asciidoctor.js since we need to use a `FileSystemProvider` https://code.visualstudio.com/api/references/vscode-api#FileSystemProvider

AshkanV:
Thank you so much.

I read your final merged code to learn more about the internal functionality of the system and also how to write the required tests.

I have only two questions.

1. Could you please describe a little more why couldn't we just use `data-uri` and needed a new attribute for it (`kroki-data-uri`)?
2. Is the new code uses any file caching?

Because I saw that no file is ever created and so every time all the diagrams should be fetched again, which really takes quite some times on big documents that normally contains no changes in most of diagrams.

I tested the behavior with the Ruby version and it also creates fetched images in the `imagesoutdir`, so maybe it is more consistent if JS version also behaves similarly and provides the file caching mechanisms.

I also test it with my original pull request, exported an HTML from the test diagram (which creates a file in the `imagesoutdir`) and then stopped the Kroki container. Then I export another HTML with no server, and the file is read from the disk and embedded in the final HML. Also if I delete/remove the cached image on the disk and retry to export, the HTML just contains a text of the PlantUML diagram and no image and no link to the server is included (because of `kroki-fetch-diagram` attribute).

Thanks

AshkanV:
If it is of any help, I also updated my code to remove extra parameter in save function and also return the full file path of the file in case of no `data-uri` attribute, And also included the test from the merged commit (with one change: `kroki-data-uri` -> `data-uri`)

ggrossetie:
> Could you please describe a little more why couldn't we just use data-uri and needed a new attribute for it (kroki-data-uri)?

We support both but in VS code `data-uri` is forcibly disabled since reading local files from VS code is not supported by Asciidoctor.js and will throw an error.
Example:

```
= Title
:data-uri:

image:path/to/foo.png[]
```

The above document used to throw an error in VS code so we've decided to forcibly disable `data-uri`.

> Is the new code uses any file caching?

No, we are reading the content and converting it to base64 in memory.

> I tested the behavior with the Ruby version and it also creates fetched images in the imagesoutdir

When using `data-uri`? That's a bit unexpected. In my opinion if you want to embed images in your HTML document, we should not write them in the `imagesoutdir`. We could however write them in a "cache directory".
This is somehow related to https://github.com/asciidoctor/asciidoctor-kroki/issues/90 and something we could implement independently of `data-uri` (or `kroki-data-uri`).

AshkanV:
Yes its is correct that `data-uri` is not supported directly in VSCode and the developer tool shows that only the image file address is there, but when we export an HTML with this attribute enalbed, it puts a base64 encoding of the image in the resulting file.

> When using data-uri? That's a bit unexpected. In my opinion if you want to embed images in your HTML document, we should not write them in the imagesoutdir. We could however write them in a "cache directory".

I also think that is correct, but what about when `data-uri` is combined with `kroki-fetch-diagram`? Shouldn't it first fetch (and save the diagram to the disk) and then encode it in the resulting document? 

Thank you

AshkanV:
> When using data-uri?

Yes, in my exprience:

This is the sample file (test.ad):
```ad
= Title
:data-uri:
:diagram-server-type: kroki_io
:diagram-server-url: https://kroki.io

[plantuml]
----
@startuml
Alice -> Bob: Authentication Request
Alice <-- Bob: Another authentication Response
@enduml
----
```

and the command to run is:
```bash
asciidoctor -r asciidoctor-diagram test.ad
```

This creates a picture file: (diag-plantuml-md5-98ade79bc4f3c15dbb15304e0097be58.png) and a HTML file with the image data base64 encoded in it.

AshkanV:
I also tested the asciidoctor-kroki directly (instead of asciidoctor-diagram) and the result is:

1. If I only use `data-uri` it only puts a link of the diagram to the server in the resulting HTML.
2. If I use both `data-uri` and `kroki-fetch-diagram` it first downloads the file to the `imagesoutdir` (or beside the document, if not set) and then base64 encode it in the resulting HTML.

This is the sample file (test.ad):

```ad
= Title
:data-uri:
:kroki-fetch-diagram:

[plantuml]
----
@startuml
Alice -> Bob: Authentication Request
Alice <-- Bob: Another authentication Response
@enduml
----
```

and the command to run:
```bash
asciidoctor -r asciidoctor-kroki test.ad
```

ggrossetie:
It makes more sense to write the disk on disk when `kroki-fetch-diagram` is defined. So it seems that Asciidoctor Diagram automatically fetches and writes files on disk?

AshkanV:
> It makes more sense to write files on disk when `kroki-fetch-diagram` is defined. So it seems that Asciidoctor Diagram automatically fetches and writes files on disk?

Yes it looks like that "Asciidoctor Diagram" does that with only `data-uri` enabled and "Asciidoctor Kroki" (Ruby version) does that when both `data-uri` and `kroki-fetch-diagram` is set.

ggrossetie:
Sorry I lost track of this pull request. Can we close it or do we still need to do something?

AshkanV:
Hello, Sorry I thought you might be busy, so I didn't bother anymore
I created a local branch for myself, which embed the diagram images if both `kroki-fetch-diagram` and `data-uri` attributes are set it fetches the diagram and save it on the disk which is used as a cache and also base64-encode it in the final output

But I see that after the changes made here, it has conflicts with current codes

PR #455, opened 2024-08-08
Fix preprocess of subfolder includes for antora

herbrechtsmeier:
The preprocessing needs to respect the folder of the current processed file because the includes are relative to this file.


ggrossetie:
It looks good to me :+1:

PR #480, opened 2026-01-12
Update Antora documentation link for Asciidoctor extensions

mmattel:
In addition updating the extension link for Antora, the statement that Antora it is not compatible with svg when using Kroki ist not longer true. I have tested this and it works flawless.


ggrossetie:
Thanks @mmattel

PR #508, opened 2026-08-01
fix: locate fetched diagram when imagesoutdir diverges from imagesdir

### LINKED ISSUE #373: Fetched diagram cannot be located if image output directory does not match imagesdir
If the image output directory does not match the value of `imagesdir`, such as when `imagesoutdir` is specified, the converter cannot locate the fetched diagram. This particularly affects Asciidoctor PDF, which must read the image in order to embed it.

The current workaround for this situation in Asciidoctor Diagram is to use an absolute path as the image target if the `data-uri` attribute is set on the document. See:

* https://github.com/asciidoctor/asciidoctor-pdf/blob/main/lib/asciidoctor/pdf/converter.rb#L139-L140
* https://github.com/asciidoctor/asciidoctor-diagram/blob/master/lib/asciidoctor-diagram/diagram_processor.rb#L238

Since setting the `data-uri` has other implications, we may decide to settle on a different attribute name.

Since we plan to add support for the `imagesdir` on the image node in core (see https://github.com/asciidoctor/asciidoctor/issues/3660), one approach is to rely on that enhancement and set the `imagesdir` attribute on the generated node if it differs from the normal location. That way, the converter doesn't have to do any extra work.

ggrossetie:
When `imagesoutdir` differs from where `imagesdir` would normally place the file, the converter (notably Asciidoctor PDF, which reads the file back off disk) couldn't find the fetched diagram: the image reference was still built from the document-wide `imagesdir`. This sets an `imagesdir` override on the generated image node instead, pointing at wherever the file was actually written, leveraging the node-level `imagesdir` lookup added in asciidoctor/asciidoctor#3660.

Effective now in the JS/Node.js extension (Asciidoctor.js 4.0+ already ships the node-level lookup). The Ruby gem carries the same fix but it stays inert until Asciidoctor Ruby releases the equivalent core change (merged upstream, no milestone yet).

Resolves #373.


mojavelinux:
Yes, that's consistent with Asciidoctor Diagram I believe.

